PR #24, opened 2018-04-17
Add flag --start_with_program for mip start

emmser:


trailblazer/cli/core.py
```diff
@@ -74,14 +74,15 @@ def log_cmd(context, sampleinfo, sacct, quiet, config):
 @click.option('-p', '--priority', type=click.Choice(['low', 'normal', 'high']), default='normal')
 @click.option('-d', '--dryrun', is_flag=True, help='only generate SBATCH scripts')
 @click.option('--command', is_flag=True, help='only show the MIP command')
+@click.option('--startfrom', help='start the pipeline from this step, see format for program using mip --help')
 @click.argument('family', required=False)
 @click.pass_context
-def start(context, mip_config, email, priority, dryrun, command, family):
+def start(context, mip_config, email, priority, dryrun, command, startfrom, family):
```
henrikstranneheim: Should you not call it "start_with_program" to mimic MIP syntax? I might add a start_from_program flag later and they imply different starting points.

trailblazer/mip/start.py
```diff
@@ -22,7 +22,8 @@
         'option': '--gatk_variantrecalibration_snv_max_gaussians',
         'default': '1',
     },
-    'skip_evaluation': {'option': '--qccollect_skip_evaluation'}
+    'skip_evaluation': {'option': '--qccollect_skip_evaluation'},
+    'startfrom': {'option': '--start_with_program'}
```
henrikstranneheim: Add a comma at the end

trailblazer/cli/core.py
```diff
@@ -74,14 +74,15 @@ def log_cmd(context, sampleinfo, sacct, quiet, config):
 @click.option('-p', '--priority', type=click.Choice(['low', 'normal', 'high']), default='normal')
 @click.option('-d', '--dryrun', is_flag=True, help='only generate SBATCH scripts')
 @click.option('--command', is_flag=True, help='only show the MIP command')
+@click.option('--startfrom', help='start the pipeline from this step, see format for program using mip --help')
```
henrikstranneheim: from this step -> beginning with program

trailblazer/cli/core.py
ingkebil: kindof weird comment. This is a wrapper, which follows the click and python syntax. All options in TB and CG are very very consice (and use `-` as delimiter ;))

trailblazer/mip/start.py
ingkebil: Why's that?

ingkebil:
Ok, my comments look weird like that. They were meant as replies to Henrik's comments.

henrikstranneheim:
@ingkebil: I am kinda of curious why my comment is weird? You are missing the point! Renaming the flag name adds complexity and reduses readability as it disconnects what it is actually wrapping. I am all for consistency so please use the appropriate delimiter. 

Regarding the comma: if you add another option later, will that not cause an error if you forget to add the comma to the line above? If so, why introduce the subtle bug instead of adding a comma directly.

ingkebil:
About the comma: my thinking was that trailing commas make a string into a tuple and add an empty element to a list. Turns out I'm wrong and adding a trailing comma to a dict makes for nicer diff's. See https://stackoverflow.com/questions/29063868/does-the-extra-comma-at-the-end-of-a-dict-list-or-set-has-any-special-meaning

ingkebil:
About the flag: aha OK. Good point. It's about documentation. It was clear from the code how we name parameters in TB.

I want to avoid having mixed syntax params seep into packages. So far we went for split the words on `-` (mostly because it reduces typing work). TB is not exactly following that and that's another problem. 

I understand and agree with your point tho, so let's go for that.

ingkebil:
Yeah exactly. It would be worth pointing out hyphened options in the style guide in the development manual: https://github.com/Clinical-Genomics/development/issues/56

That being said, TB is older than `cg` and the options are named less consistently (e.g. `--log-level` but `--dryrun`).

Henrik's comment is about setting options that are going to be set for another pipeline. So there it does make sense for consistency to name the options identical in both tools, even if this means deviating from our proposed norm (being: use the `-`).
**That the option deviates from the norm, should be pointed out in a comment in the code.**

hassanfa:
Of course! But as long as it doesn't create parallel local standards. Sacrificing standards for the sake of consistency will slowly start introducing unpredictable coding styles.

emmser:
I have made some changes after your comments, please take a look

ingkebil:
Looking good!

PR #25, opened 2018-04-25
rewriting_trending_from_mip6_into_mip5 for delivery reports

patrikgrenfeldt:
…azer

### DIFF
--- a/trailblazer/mip/trending.py
+++ b/trailblazer/mip/trending.py
@@ -3,6 +3,7 @@
 
 from trailblazer.mip import files
 
+
 def parse_mip_analysis(mip_config_raw: dict, qcmetrics_raw: dict, sampleinfo_raw: dict) -> dict:
     """Parse the output analysis files from MIP for adding info
     to trend database
@@ -46,18 +47,17 @@ def parse_mip_analysis(mip_config_raw: dict, qcmetrics_raw: dict, sampleinfo_raw
 
     ## Add sample level info from qc_metric file
     for sample_data in qcmetrics_data['samples']:
-
         ## Add duplicate reads
         duplicates_percent = sample_data['duplicates'] * 100
         outdata['duplicates'][sample_data['id']] = duplicates_percent
 
         ## Add at and gc droputs
-        #outdata['at_dropout'][sample_data['id']] = sample_data['at_dropout']
-        #outdata['gc_dropout'][sample_data['id']] = sample_data['gc_dropout']
+        # outdata['at_dropout'][sample_data['id']] = sample_data['at_dropout']
+        # outdata['gc_dropout'][sample_data['id']] = sample_data['gc_dropout']
 
         ## Add insert size metrics
-        #outdata['median_insert_size'][sample_data['id']] = sample_data['median_insert_size']
-        #outdata['insert_size_standard_deviation'][sample_data['id']] = sample_data[
+        # outdata['median_insert_size'][sample_data['id']] = sample_data['median_insert_size']
+        # outdata['insert_size_standard_deviation'][sample_data['id']] = sample_data[
         # 'insert_size_standard_deviation']
 
         ## Add mapped reads
--- a/trailblazer/mip/trending.py
+++ b/trailblazer/mip/trending.py
@@ -19,14 +19,10 @@ def parse_mip_analysis(mip_config_raw: dict, qcmetrics_raw: dict, sampleinfo_raw
     ### Define output dict
     outdata = {
         'analysis_sex': {},
-        'at_dropout': {},
         'family': None,
         'duplicates': {},
-        'gc_dropout': {},
         'genome_build': None,
-        'insert_size_standard_deviation': {},
         'mapped_reads': {},
-        'median_insert_size': {},
         'mip_version': None,
         'sample_ids': [],
     }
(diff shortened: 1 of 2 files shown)


trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        qcmetrics_raw (dict): raw YAML input from MIP analysis qc metric file
+        sampleinfo_raw (dict): raw YAML input from MIP analysis qc sample info file
+    Returns:
+        dict: parsed data
+    """
+
+    ### Define output dict
+    outdata = {
+        'analysis_sex': {},
+        'at_dropout': {},
```
ingkebil: We could comment this one out

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        'family': None,
+        'duplicates': {},
+        'gc_dropout': {},
```
ingkebil: This one as well

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        'genome_build': None,
+        'insert_size_standard_deviation': {},
```
ingkebil: And this one

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        'mapped_reads': {},
+        'median_insert_size': {},
```
ingkebil: And this one

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Args:
+        mip_config_raw (dict): raw YAML input from MIP analysis config file
+        qcmetrics_raw (dict): raw YAML input from MIP analysis qc metric file
+        sampleinfo_raw (dict): raw YAML input from MIP analysis qc sample info file
+    Returns:
+        dict: parsed data
+    """
+
+    ### Define output dict
```
ingkebil: What's up with these triple `#`?

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        'gc_dropout': {},
+        'genome_build': None,
+        'insert_size_standard_deviation': {},
+        'mapped_reads': {},
+        'median_insert_size': {},
+        'mip_version': None,
+        'sample_ids': [],
+    }
+    ### Config
+    ## Parse raw mip_config into dict
```
ingkebil: What's up with these dual `#`?

trailblazer/mip/trending.py
```diff
@@ -0,0 +1,80 @@
...
+        outdata['analysis_sex'][sample_data['id']] = sample_data['predicted_sex']
+
+    ### Qc sample info
+    ## Parse qc sample info file
+    sampleinfo_data = files.parse_sampleinfo(sampleinfo_raw)
+
+    ## Add mip version
+    outdata['mip_version'] = sampleinfo_data['version']
+
+    ## Add mip version
```
ingkebil: `# Add genome build`

trailblazer/mip/trending.py
henrikstranneheim: This is an important qc-metric to trend as was decided previously. Are you adding this info from some other source?
I think it is bad practise to leave lines of code that are not used in a "commented form". It clutteres the code and makes it more complicated than it is. Removing the unused line is prefered and then adding it in when required. In this case I do not really understand why you want to comment it out as you can just ignore this key when parsing the output in the downstream process (if it is not ready for the "at_dropout"), but I am not up to speed on what you are planning here.

trailblazer/mip/trending.py
henrikstranneheim: This is an important qc-metric to trend as was decided previously. Are you adding this info from some other source?

trailblazer/mip/trending.py
henrikstranneheim: This is an important qc-metric to trend as was decided previously. Are you adding this info from some other source?

trailblazer/mip/trending.py
henrikstranneheim: This is an important qc-metric to trend as was decided previously. Are you adding this info from some other source?

trailblazer/mip/trending.py
henrikstranneheim: It's just a way of commenting what the next block of code is handling. If you have another system in place in trailblazer feel free to modify accordingly.

trailblazer/mip/trending.py
henrikstranneheim: see above

trailblazer/mip/trending.py
henrikstranneheim: Ok, now I see you are just trying to get it to be compatible with the mip5.0 in production, right?

ingkebil: ... like the title said ;)

patrikgrenfeldt:
All requested changes performed, what do you think of it now?

henrikstranneheim:
Wow, You really went all in for utility subs, 😁. Very nice and clear! 👍

ingkebil:
Let's merge and :ship:

trailblazer/mip/trending.py
ingkebil: Python has a style guide which we are following: https://www.python.org/dev/peps/pep-0008/
e.g. Pycharm automatically highlights any deviations from pep-8.

PR #26, opened 2018-04-26
Updated name of option startwith to include '-' as a separator of words

emmser:


ingkebil:
Looking good! :+1:

ingkebil:
I've created a clean branch which is under PR in #27 . It's the same commits you made (even authored by you!) but without the mip6 branch commits.

I will close this commit. Hope that's fine.

PR #35, opened 2018-09-14
Initial commit pipeline interface

ingkebil:
This was the initial work we put in beginning of March. To summarize: it makes the interface for starting a pipeline more generic instead of having it MIP specific. It also defines the interface or provides a template on how to define the options for a pipeline.

This code is untested and requires unit tests and a manual check that it would actually work.

Feel free to improve, work on top of this, or comment.


ingkebil:
Yes, this will be moved to `cg` instead.

PR #45, opened 2019-02-22
Continues when mip data is missing

patrikgrenfeldt:
This PR makes trailblazer get data from mip output without failing as soon as the structure is different

How to test:
1. install on rasta stage
1. us
1. run `cg upload delivery-report -p [custid] [familyid]` on an somewhat older case that fails in production with `KeyError`

Expected output:
A delivery report

### DIFF
--- a/trailblazer/mip/files.py
+++ b/trailblazer/mip/files.py
@@ -1,9 +1,17 @@
 # -*- coding: utf-8 -*-
 """Parse the MIP config file."""
-from typing import TextIO
 import csv
+from typing import TextIO
 
 PED_SEX_MAP = {1: 'male', 2: 'female', 0: 'unknown'}
+
+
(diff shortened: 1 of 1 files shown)


ingkebil:
Did you just do a search and replace here or is there a difference between mandatory elements and others?

patrikgrenfeldt:
@ingkebil, @emililo do we want this in production or do we want to have the
code failing so we know when the code and data don't match?

Den sön 24 feb. 2019 21:47Kenny Billiau <notifications@github.com> skrev:

ingkebil:
I just discussed this with @emiliaol and we decided not to push this to production.

I think it's a good idea to pursue this idea and to create a parser for each version of MIP at some point.

patrikgrenfeldt:
Closing since we don't want this behavior in production

PR #70, opened 2019-10-10
Show failed last month

patrikgrenfeldt:
This PR adds functionality to show stats only last 31 days

**How to install**:
1. install on stage of the coffee machine: `bash servers/resources/clinical-db.scilifelab.se/update-trailblazer-ui-stage.sh THISBRANCH`
1. activate stage: `us`

**How to test**:
1. log into tb-stage
1. press stats

**Expected outcome**:
`tb` should shows stats for the last 31 days only.
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by @ingkebil 
- [ ] tests executed by @patrikgrenfeldt 
- [ ] "Merge and deploy" approved by @ingkebil 

This is |minor| **version bump** because it changes functionality in a backwards compatible manner

### DIFF
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -75,10 +75,10 @@ def me():
     return jsonify(**g.current_user.to_dict())
 
 
-@blueprint.route('/aggregate/jobs/<int:days_back>', methods=['GET'])
-def aggregate_jobs(days_back: int = 31):
+@blueprint.route('/aggregate/jobs')
+def aggregate_jobs():
     """Return stats about jobs."""
-
+    days_back = int(request.args.get('days_back', 31))
     one_month_ago = datetime.datetime.now() - datetime.timedelta(days=days_back)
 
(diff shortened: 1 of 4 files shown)


patrikgrenfeldt:
Test result:
image

trailblazer/server/api.py
```diff
@@ -76,5 +78,8 @@ def me():
 @blueprint.route('/aggregate/jobs')
 def aggregate_jobs():
```
ingkebil: we could even give it a param that defaults to 31. Then we can surf to `/aggregate/jobs/60` and see more ;)

I don't know where this is used, so I don't know if this makes sense.

ingkebil:
Approved

trailblazer/server/api.py
patrikgrenfeldt: I think there is more to it, it's not the address of the backend that we use in the browser so we need to relay what is written there through the Vue application

Issue #239, opened 2023-07-17
Error with job status


fevac:
Trailblazer is faling when status in the job table is "canceled". This is fixed when changed to "cancelled".

Currently, status in the analysis table and the job table ARE NOT THE SAME and they accept different values.
```
class Analysis(Model):
  status = Column(types.Enum(*STATUS_OPTIONS))
('pending', 'running', 'completed', 'failed', 'error', 'canceled', 'completing', 'qc')
```

```
class Job(Model):
    status = Column(types.Enum(*JOB_STATUS_OPTIONS))
('completed', 'running', 'pending', 'completing', 'failed', 'cancelled', 'timeout')
```

Both lists are very similar but not identical.

Tower is currently using the same status values for both analysis and job tables and this leads to the error above. 

- Quick fix:  not allow `canceled` or `error` and return `failed` (status accepted by both tables) instead.

- Permanent fix: standarize status in both tables, as there are only few differences: cancelled/canceled, timeout, error, (qc). We should use `cancelled` instead of `canceled`

fevac:
@henrikstranneheim any objection against making the allowed values for Job Status the same as Analysis status?? that would require making some changes so values for SlurmJobStatus are converted to TrailblalzerStatus  prior storing in the sql database

henrikstranneheim:
> @henrikstranneheim any objection against making the allowed values for Job Status the same as Analysis status?? that would require making some changes so values for SlurmJobStatus are converted to TrailblalzerStatus prior storing in the sql database

I think that sounds like the way to go! @seallard If you have time - it would be great to fix this

seallard:
For the refinement:
- Do we need to transform the existing data in the status column to ensure all statuses are set to valid values?
- What should the new statuses be called?
- Do any other repos rely on any of these status fields and need to be updated as well?

henrikstranneheim:
> For the refinement:
> 
>     * Is there no semantic value in distinguishing the status for a job and an analysis? Are they always the same?

No I do not think so. Analysis can be a subset, but canceled should be fixed to cancelled.
> 
>     * What should the new status enum be called? TrailblazerStatus?

Yes, it already exist, together with SlurmJobStatus (for Jobs).
> 
>     * Do we need to transform the existing data in the status column to ensure all statuses are set to valid values according to the new enum?
> 
Yes, they should be aligned.

>     * Do any other repos rely on any of these status fields and need to be updated as well?

Might be in Cg and Cigrid

[...]

PR #296, opened 2023-11-06
Fix enum spelling

seallard:
## Description
First step of resolving https://github.com/Clinical-Genomics/trailblazer/issues/239. This PR fixes the spelling of the analysis.status enum field `canceled` to `cancelled`.

Worst case, a status will change color in cigrid-ui after this patch from what I understand.

### Fixed
- Change spelling of enum field
- Add revision for the change to reflect it in the database

## Testing
Applied revision to stage, looks ok.


islean (approved):
Looks beautiful!

PR #299, opened 2023-11-10
Replace Alchy with SQLAlchemy

seallard:
## Description
This PR replaces `Alchy` with `SQLAlchemy`. `Alchy` is not maintained and locks us to an old version of `SQLAlchemy`, which is why we want to replace it. Closes https://github.com/Clinical-Genomics/trailblazer/issues/298.

The session/transaction management approach applied here ties the session and transaction scope to that of a request or CLI command. One session and one transaction is used per request/command. 

Once the request/command has been processed, the transaction is committed and the session is removed (unless an exception occurred - then the transaction is rolled back first). I think it makes sense since it is very simple and the requests/commands run quickly (as far as I understand).

This differs from what is done in CG where we limit the scope of the transactions more tightly by doing commits throughout the codebase. I think that approach makes sense in cg since some of the commands run for 10 + hours. We don't want to keep a single transaction open for that long, it would be problematic since transactions can lock tables.

TODOs:
- [x] Add database module
- [x] Replace `Flask-Alchy`
- [x] Replace `AlchyManager`
- [x] Replace `Alchy` in models with `SQLAlchemy`
- [x] Add session/transaction management for CLI and API
- [x] Use global `scoped_session` in handlers
- [x] Remove commits in codebase
- [x] Patch tests
- [x] Add serialization (`Alchy` had built in `to_dict` methods on models)

I'd appreciate some help with the QA testing of this PR.

### Fixed
- Refactor project to only use SQLAlchemy
- Standardize session/transaction management
- Remove Flask-Alchy
- Remove Flask-SQLAlchemy
- Upgrade Click to latest version
- Upgrade Flask to latest version

```
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_trailblazer -t trailblazer -b  remove-alchy -a
```

### This [version](https://semver.org/) is a
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

PR #306, opened 2023-11-15
Commit transactions before sending response

seallard:
## Description
The transaction management approach taken in https://github.com/Clinical-Genomics/trailblazer/pull/299 is not ideal since the transactions are committed only after responses are sent. This PR restores the original commits in the CRUD handlers.

By committing before we return a response, we ensure that the caller receives an error if something goes wrong against the database.

### Fixed
- Commit transactions before sending response.

### How to test
- [ ] Ensure cigrid views work and data is written
- [ ] Ensure cli works and data is written

### DIFF
--- a/trailblazer/store/crud/delete.py
+++ b/trailblazer/store/crud/delete.py
@@ -20,6 +20,7 @@ class DeleteHandler(BaseHandler):
         session: Session = get_session()
         for job in analysis.jobs:
             session.delete(job)
+        session.commit()
 
     def delete_analysis(self, analysis_id: int, force: bool = False) -> None:
         """Delete the analysis from the database. Also delete ongoing analysis if 'force' is True.
@@ -37,3 +38,4 @@ class DeleteHandler(BaseHandler):
             )
         LOG.info(f"Deleting analysis: {analysis_id}, for case: {analysis.family}")
         session.delete(analysis)
+        session.commit()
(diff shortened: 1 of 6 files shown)


diitaz93 (approved):
LGTM 👍

trailblazer/store/crud/delete.py
```diff
@@ -20,6 +20,7 @@ def delete_analysis_jobs(self, analysis: Analysis) -> None:
         session: Session = get_session()
         for job in analysis.jobs:
             session.delete(job)
+        session.commit()
```
Vince-janv, line 23: Do we want this here, or should we perform the commit in the CLI/Endpoint level? If possible I would prefer to have it on the same level for all

Vince-janv (approved):
Nice 👍

trailblazer/store/crud/delete.py
seallard: Me too, but I don't like cluttering the cli and api with these commits and I don't want to spend more time thinking about it :sweat_smile: I'm leaving it for now as is, if someone (probably me in the future) feel like they have time and energy to refactor it

seallard: The session/transaction management should not be done in the controller or CLI layer either. I think one way to go is to use the `teardown_request` decorator https://flask.palletsprojects.com/en/3.0.x/api/#flask.Flask.teardown_request. But I'm pretty over dealing with this. Feel free to pick it up if you are motivated :muscle:

seallard:
Views work, fields can be written. CLI works, scan and write.

Issue #309, opened 2023-11-15
Speed up loading analyses


seallard:
The `/analyses` endpoint is very slow, especially for sars-cov-2.

seallard:
Slight speedup after Python upgrade
![image](https://github.com/Clinical-Genomics/trailblazer/assets/33596072/e9de4c8b-bc74-42fa-b971-7a654d39258b)

seallard:
Filtering is fast, as is re-ordering and searching. The `search_query` param is only used to provide the pipeline. The query above is only used in the `/analyses` endpoint. So... This can be sped up a lot easily 🚀

seallard:
Removing the jobs from the response did it 🚀 
![image](https://github.com/Clinical-Genomics/trailblazer/assets/33596072/d6aeeffd-96e7-4cb4-9e21-fb28a1370153)

PR #311, opened 2023-11-15
Speedup retrieving analyses

seallard:
## Description
This PR speeds up the `/analyses` endpoint :zap:  It used to be :turtle: and now it is more like :zebra:  Closes https://github.com/Clinical-Genomics/trailblazer/issues/309.

From 25 seconds to 100 ms for retrieving the sars-cov data (rendering is still slow, but that is a frontend issue).
 
The response size was very large (like 70 MB for sars-cov-2), excluding the jobs takes it down to around 200 kB tops. The jobs are not needed in the analysis table in cigrid, so there is no reason to include them in the `/analyses` response. They are fetched again from `/analyses/<id>` when you click on an analysis to get the detailed view.

I have not found any other repo that calls the `/analyses` endpoint.

### Fixed
- Speed up retrieving analyses per pipeline

### How to test
- [x] Deploy to stage vm and check that the views work and load faster

### DIFF
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -65,7 +65,6 @@ def analyses():
     for analysis in query_page.all():
         analysis_data = analysis.to_dict()
         analysis_data["user"] = analysis.user.to_dict() if analysis.user else None
-        analysis_data["jobs"] = [job.to_dict() for job in analysis.jobs]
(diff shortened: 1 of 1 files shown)


Vince-janv (approved):
Elegant! 🎩

PR #319, opened 2023-11-27
Patch update endpoint

seallard:
## Description
The endpoint parsed the `is_visible` field as a string instead of as a bool, causing the hide functionality to break in cigrid-ui. This PR adds a pydantic model for parsing the analysis update request. Closes https://github.com/Clinical-Genomics/cigrid-ui/issues/458.

It would be nice to set up some tests that call the actual endpoints. That way, these bugs would not slip through.

### Fixed
- Toggling of analyses visibility

### How to test
- [ ] deploy to stage
- [ ] ensure analysis stays hidden after refreshing page

### DIFF
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -6,6 +6,8 @@ from typing import Mapping
 
 from flask import Blueprint, Response, abort, g, jsonify, make_response, request
 from google.auth import jwt
+from http import HTTPStatus
+from pydantic import ValidationError
 from sqlalchemy.orm import Query
 
 from trailblazer.constants import (
@@ -14,6 +16,7 @@ from trailblazer.constants import (
     TrailblazerStatus,
 )
 from trailblazer.server.ext import store
+from trailblazer.server.schemas import AnalysisUpdate
 from trailblazer.store.models import Analysis, Info, User
 from trailblazer.utils.datetime import get_date_number_of_days_ago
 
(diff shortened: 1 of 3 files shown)


seallard:
Deployed and tested in stage, analyses are now hidden as expected.

trailblazer/server/api.py
```diff
@@ -77,12 +80,16 @@ def analysis(analysis_id):
...
-        store.update_analysis(
-            analysis_id=analysis_id, comment=comment, status=status, is_visible=is_visible
-        )
+        try:
+            analysis_update = AnalysisUpdate.model_validate(request.json)
+            store.update_analysis(
+                analysis_id=analysis_id,
+                comment=analysis_update.comment,
+                status=analysis_update.status,
+                is_visible=analysis_update.is_visible,
```
seallard, line 89: Might be cleaner to pass the entire schema here instead to the `update_analysis` method. Let me know what you think.

trailblazer/server/schemas.py
```diff
@@ -0,0 +1,7 @@
+from pydantic import BaseModel
+
+
+class AnalysisUpdate(BaseModel):
+    is_visible: bool | None = None
+    comment: str | None = None
+    status: str | None = None
```
seallard, line 7: Maybe this field should be typed with the status enum type. Let me know what you think.

trailblazer/server/api.py
```diff
@@ -77,12 +80,16 @@ def analysis(analysis_id):
...
+            )
+        except ValidationError as e:
```
henrikstranneheim, line 91:
```suggestion
        except ValidationError as error:
```

trailblazer/server/api.py
henrikstranneheim: I think this is nice since it validates the request 👍

henrikstranneheim (approved):
Great 👍

trailblazer/server/schemas.py
henrikstranneheim: Good point 👍

Issue #321, opened 2023-11-28
Slow processing of request to /analyses endpoint


seallard:
After re-adding that the most recent failed job is added to the response in the `/analysis` endpoint, we are getting slow response times (not as slow as it used to be though). This time for a different reason, the response size is reasonable - but the server is taking forever processing the request

![image](https://github.com/Clinical-Genomics/trailblazer/assets/33596072/44355477-4a2e-4310-8420-125f963fca40)

seallard:
For covid, we have on average 1000 jobs per analysis. When we retrieve 300 analysis at once, we are for each analysis

1) Filtering the job list on status
2) Sorting it on date

This is likely the bottleneck which causes the slow response time. The property should run a more efficient query.

PR #322, opened 2023-11-28
Retrieve latest failed job via query

seallard:
## Description
This PR speeds up the retrieval of the latest failed job for an analysis. Closes https://github.com/Clinical-Genomics/trailblazer/issues/321.


trailblazer/server/api.py
```diff
@@ -68,7 +68,7 @@ def analyses():
     for analysis in query_page.all():
         analysis_data = analysis.to_dict()
         analysis_data["user"] = analysis.user.to_dict() if analysis.user else None
-        failed_job: Job | None = analysis.last_failed_job
+        failed_job: Job | None = store.get_latest_failed_job_for_analysis(analysis.id)
```
seallard, line 71: This is not super efficient (running one extra query per analysis) but way better than the previous implementation. Created an issue for us to improve it https://github.com/Clinical-Genomics/trailblazer/issues/324

tests/store/filters/test_job_filters.py
```diff
@@ -52,3 +52,17 @@ def test_filter_jobs_by_status(job_store: MockStore):
...
     # THEN the jobs attribute status should match the original
     assert jobs[0].status == TrailblazerStatus.COMPLETED
+
+
+def test_filter_jobs_by_analysis_id(job_store: MockStore):
+    """Test return jobs by analysis id."""
+    # GIVEN a store containing jobs
+
+    # WHEN retrieving a job by analysis id
+    jobs: Query = job_store.get_query(table=Job).filter_by(analysis_id=1)
```
islean, line 62: Do we not want to test `filter_jobs_by_analysis_id` here? A bit unsure what this tests.

tests/store/crud/test_read.py
```diff
@@ -272,3 +272,16 @@ def test_get_users_no_username(user_store: MockStore, user_email: str):
 
     # THEN the user should be returned
     assert users[0].email == user_email
+
+
+def test_get_latest_failed_job_for_analysis(job_store: MockStore):
```
islean, line 277: Would it be possible to test other behaviours like what happens when an analysis doesn't have a failed job? And potentially preferable to use a real Store, rather than a mock store.

islean (approved):
Looks good! Would love to see more tests to make any future changes safer, but I guess there is always a balance that needs to be struck with time.

tests/store/filters/test_job_filters.py
seallard: Ah, haha did not question the copilot enough

tests/store/crud/test_read.py
seallard: Added it :+1:

PR #326, opened 2023-12-08
Ensure the status is updated for the correct analysis

seallard:
Closes https://github.com/Clinical-Genomics/cigrid-ui/issues/468. When the analysis status was updated a function called `update_analysis_status` was called. This function does update the status, but it does it for the most recent analysis for a case 🐛 🥲 Patched it so we just set the status directly on the analysis.

I'll set up a separate PR adding some integration tests for our endpoints.

### Fixed
- Updating the status on an analysis

Tested in stage.

### DIFF
--- a/trailblazer/store/crud/update.py
+++ b/trailblazer/store/crud/update.py
@@ -264,7 +264,7 @@ class UpdateHandler(BaseHandler):
             analysis.is_visible = bool(is_visible)
 
         if status:
-            LOG.info(f"Setting status to {status} for analysis {analysis.family}")
+            LOG.info(f"Setting status to {status} for analysis {analysis.id}")
             analysis.status = status
 
         session: Session = get_session()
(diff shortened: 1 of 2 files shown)


trailblazer/store/crud/update.py
```diff
@@ -256,15 +256,16 @@ def update_analysis(
         analysis: Analysis = self.get_analysis_with_id(analysis_id)
 
         if comment:
-            LOG.info(f"Adding comment {comment} to analysis {analysis.family}")
+            LOG.info(f"Adding comment {comment} to analysis {analysis.id}")
```
islean, line 259: Do we have anything better in Trailblazer to identify analyses with than their integer id?

islean (approved):
Looks good!

trailblazer/store/crud/update.py
seallard: Well, using the case name it is associated with is misleading since we have many analyses per case. I think the database id for the case is the most reasonable identifier at the moment

PR #334, opened 2024-01-02
Include hidden analyses when searching

seallard:
## Description
The expected behaviour is that analyses which have been hidden are included when searching.


Vince-janv (approved):
Short and sweet 👍

PR #339, opened 2024-01-08
Improve logging for scan command

seallard:
## Description
Improve logging for failing scan command, see https://github.com/Clinical-Genomics/trailblazer/issues/340.

The scan command sometimes fails with a `FileNotFound` error. This is most likely due to the slurm job file missing for some reason. However, the log message currently does not include the file path so it is hard to tell which file is missing.

### Fixed
- Improve logging for scan command

### DIFF
--- a/trailblazer/apps/slurm/api.py
+++ b/trailblazer/apps/slurm/api.py
@@ -32,12 +32,10 @@ def cancel_slurm_job(slurm_id: int, analysis_host: str | None = None) -> None:
 
 def get_slurm_squeue_output(slurm_job_id_file: Path, analysis_host: str | None = None) -> str:
     """Return squeue output from ongoing analyses in SLURM."""
-    try:
-        slurm_job_id_file_content: dict[str, list[str]] = ReadFile.get_content_from_file(
-            file_format=FileFormat.YAML, file_path=slurm_job_id_file
-        )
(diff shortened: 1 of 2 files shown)


trailblazer/apps/slurm/api.py
```diff
@@ -32,9 +32,12 @@ def cancel_slurm_job(slurm_id: int, analysis_host: str | None = None) -> None:
...
     """Return squeue output from ongoing analyses in SLURM."""
-    slurm_job_id_file_content: dict[str, list[str]] = ReadFile.get_content_from_file(
-        file_format=FileFormat.YAML, file_path=slurm_job_id_file
-    )
+    try:
+        slurm_job_id_file_content: dict[str, list[str]] = ReadFile.get_content_from_file(
+            file_format=FileFormat.YAML, file_path=slurm_job_id_file
+        )
+    except FileNotFoundError as e:
+        raise FileNotFoundError(f"File not found: {slurm_job_id_file}") from e
```
ChrOertlin, line 40:
```suggestion
    except FileNotFoundError as error:
        raise FileNotFoundError(f"File not found: {slurm_job_id_file}") from error
```

ChrOertlin (approved):
Looks good! :)

Issue #342, opened 2024-01-09
Searching for certain cases gives error response


seallard:
As reported by @karlnyr and @ChrOertlin, when sending requests searching for a specific analysis it sometimes causes a 400 error response with the following message

`
1 validation error for AnalysisResponse\nanalyses.5.failed_job.started_at\n Input should be a valid datetime [type=datetime_type, input_value=None, input_type=NoneType]\n For further information visit https://errors.pydantic.dev/2.5/v/datetime_type
`

### Reproduce bug
Search for `vitalsquid` for example.

PR #343, opened 2024-01-09
Fix error response retrieving certain cases

seallard:
## Description
Closes https://github.com/Clinical-Genomics/trailblazer/issues/342. Apparently, the `started_at` field for a job can be empty.

### Fixed
- Bug where searching for specific cases causes an error response

### How to test
- [x] deploy to stage
- [x] search for vitalsquid (not present in stage... Well this fix should work)

- [ ] **PATCH**

### DIFF
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -107,9 +107,8 @@ def me():
 @blueprint.route("/aggregate/jobs")
 def aggregate_jobs():
     """Return stats about failed jobs."""
-    time_window: datetime = get_date_number_of_days_ago(
-        number_of_days_ago=int(request.args.get("days_back", ONE_MONTH_IN_DAYS))
-    )
+    number_of_days_ago: int = int(request.args.get("days_back", ONE_MONTH_IN_DAYS))
+    time_window: datetime = get_date_number_of_days_ago(number_of_days_ago)
     failed_jobs: list[dict[str, str | int]] = store.get_nr_jobs_with_status_per_category(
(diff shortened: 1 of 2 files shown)


trailblazer/server/api.py
```diff
@@ -108,9 +107,8 @@ def me():
 @blueprint.route("/aggregate/jobs")
 def aggregate_jobs():
     """Return stats about failed jobs."""
-    time_window: datetime = get_date_number_of_days_ago(
-        number_of_days_ago=int(request.args.get("days_back", ONE_MONTH_IN_DAYS))
-    )
+    number_of_days_ago: int = int(request.args.get("days_back", ONE_MONTH_IN_DAYS))
```
islean, line 110: Does this count as an instantiation of an `int` in that we can remove the type hint 🤔 Either way, you can disregard this comment

trailblazer/server/api.py
```diff
@@ -186,10 +184,8 @@ def delete(analysis_id):
 def post_get_latest_analysis():
     """Return latest analysis entry for specified case id."""
     post_request: Response.json = request.json
```
islean, line 186: How come the `post_request` is a `Response.json` object?

islean (approved):
Love all the walruses 🦭 (closest I could find)

trailblazer/server/api.py
seallard: Dunno, should be Any or None. Removing it for now

PR #344, opened 2024-01-09
Refactor API layer

seallard:
## Description
This PR refactors some of the endpoints in the web api following the pattern

Going down from the api layer to the data access layer
request -> api layer -> pydantic dto model -> service layer -> data access layer -> database model

And then back up again...
database model -> service layer -> pydantic dto model -> api layer -> response

The upside of this pattern is
- Better separation of concerns
- Get rid of logic in api layer
- Avoids passing around dicts
- Better validation of data in requests and responses

### Fixed
- Refactor API layer

### DIFF
--- a/trailblazer/dto/__init__.py
+++ b/trailblazer/dto/__init__.py
@@ -1,0 +1,6 @@
+from trailblazer.dto.analyses_request import AnalysesRequest
+from trailblazer.dto.analysis_response import AnalysisResponse
+from trailblazer.dto.failed_jobs_request import FailedJobsRequest
+from trailblazer.dto.failed_jobs_response import FailedJobsResponse
+from trailblazer.dto.analysis_update_request import AnalysisUpdateRequest
+from trailblazer.dto.analyses_response import AnalysesResponse
--- a/trailblazer/dto/analyses_response.py
+++ b/trailblazer/dto/analyses_response.py
@@ -4,22 +4,12 @@ from datetime import datetime
 
 class Job(BaseModel):
     analysis_id: int
-    context: str | None = None
     elapsed: int
     id: int
     name: str
     slurm_id: int
     started_at: datetime | None = None
     status: str
-
-
-class User(BaseModel):
-    avatar: str | None = None
-    created_at: datetime
-    email: str
-    id: int
-    is_archived: bool
-    name: str
 
 
 class Analysis(BaseModel):
@@ -41,7 +31,6 @@ class Analysis(BaseModel):
     ticket_id: str | None = None
     type: str | None = None
     uploaded_at: datetime | None = None
-    user: User | None = None
     user_id: int | None = None
     version: str | None = None
     workflow_manager: str
--- a/trailblazer/dto/analyses_response.py
+++ b/trailblazer/dto/analyses_response.py
@@ -47,6 +47,6 @@ class Analysis(BaseModel):
     workflow_manager: str
 
 
-class AnalysisResponse(BaseModel):
+class AnalysesResponse(BaseModel):
     analyses: list[Analysis]
     total_count: int
--- a/trailblazer/dto/analyses_response.py
+++ b/trailblazer/dto/analyses_response.py
@@ -2,7 +2,7 @@ from pydantic import BaseModel
 from datetime import datetime
 
 
-class FailedJob(BaseModel):
+class Job(BaseModel):
     analysis_id: int
     context: str | None = None
     elapsed: int
@@ -28,7 +28,7 @@ class Analysis(BaseModel):
     completed_at: datetime | None = None
     config_path: str | None = None
     data_analysis: str | None = None
-    failed_job: FailedJob | None = None
+    failed_job: Job | None = None
     id: int
     is_deleted: bool = False
     is_visible: bool = True
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -17,7 +17,7 @@ from trailblazer.dto.analysis_request import AnalysisRequest
 from trailblazer.dto.analysis_response import AnalysisResponse
 from trailblazer.server.ext import store
 from trailblazer.server.schemas import AnalysisUpdateRequest
-from trailblazer.server.utils import parse_analysis_request
+from trailblazer.server.utils import parse_analysis_request, stringify_timestamps
 from trailblazer.services.analysis_service import AnalysisService
 from trailblazer.store.models import Analysis, Info, User
 from trailblazer.utils.datetime import get_date_number_of_days_ago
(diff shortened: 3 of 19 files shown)


trailblazer/dto/__init__.py
```diff
@@ -0,0 +1,6 @@
+from trailblazer.dto.analyses_request import AnalysesRequest
```
henrikstranneheim, line 1: I do not know if these implicit imports in ini.py are a pattern we want? Would be nice to discus at a code review session.

trailblazer/dto/analyses_response.py
```diff
@@ -0,0 +1,52 @@
+from pydantic import BaseModel
+from datetime import datetime
+
+
+class Job(BaseModel):
+    analysis_id: int
+    context: str | None = None
```
henrikstranneheim, line 7: What context is this?

trailblazer/dto/analyses_response.py
```diff
@@ -0,0 +1,52 @@
...
+    elapsed: int
+    id: int
+    name: str
+    slurm_id: int
+    started_at: datetime | None = None
+    status: str
+
+
+class User(BaseModel):
+    avatar: str | None = None
```
henrikstranneheim, line 17: Is this used? Can we remove it?

trailblazer/server/api.py
```diff
@@ -56,39 +57,38 @@ def before_request():
 
 @blueprint.route("/analyses")
 def analyses():
```
henrikstranneheim, line 59:
```suggestion
def get_analyses():
```

trailblazer/server/api.py
```diff
@@ -56,39 +57,38 @@ def before_request():
 
 @blueprint.route("/analyses")
```
henrikstranneheim, line 58:
```suggestion
@blueprint.route("/analyses", methods=["GET"])
```

trailblazer/server/api.py
```diff
@@ -105,14 +105,14 @@ def me():
 
 
 @blueprint.route("/aggregate/jobs")
```
henrikstranneheim, line 107:
```suggestion
@blueprint.route("/aggregate/jobs", methods=["GET"])
```

trailblazer/services/analysis_service.py
```diff
@@ -8,21 +14,34 @@ class AnalysisService:
...
-        response: AnalysisResponse = self._format_response(
+        return self.create_analyses_response(
             analyses=analyses, total_analysis_count=total_analysis_count
         )
-        return response
 
-    def _format_response(
+    def get_analysis(self, analysis_id: int) -> AnalysisResponse:
+        if not (analysis := self.store.get_analysis_with_id(analysis_id)):
+            raise MissingAnalysis(f"Analysis with id {analysis_id} not found")
```
henrikstranneheim, line 25:
```suggestion
            raise MissingAnalysis(f"Analysis with id: {analysis_id} not found")
```

trailblazer/services/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+def create_jobs_response(failed_job_statistics: list[dict]) -> FailedJobsResponse:
+    return FailedJobsResponse(jobs=failed_job_statistics)
+
+
+def create_analysis_response(analysis: Analysis) -> AnalysisResponse:
+    analysis_data: dict = analysis.to_dict()
+    analysis_data["jobs"] = [job.to_dict() for job in analysis.jobs]
+    analysis_data["user"] = analysis.user.to_dict() if analysis.user else None
+    return AnalysisResponse(**analysis_data)
```
henrikstranneheim, line 13:
```suggestion
    return AnalysisResponse.model_validate(analysis_data)
```

henrikstranneheim (approved):
Beautiful 👍 . Should we take the opportunity to add unit tests for the changed end-points?

trailblazer/dto/__init__.py
seallard: Yes, I think it is a good topic 👍 The upside is that we can do the imports like this
```python
from trailblazer.dto import (
    AnalysesRequest,
    AnalysisUpdateRequest,
    AnalysesResponse,
    AnalysisResponse,
)
```
Instead of having to specify the full paths like this
```python
from trailblazer.dto.analysis_response import AnalysisResponse
from trailblazer.dto.failed_jobs_request import FailedJobsRequest
from trailblazer.dto.failed_jobs_response import FailedJobsResponse
from trailblazer.dto.analysis_update_request import AnalysisUpdateRequest
from trailblazer.dto.analyses_response import AnalysesResponse
```

I like the first option because it is more concise, and modules using another module often import multiple things from it.

seallard: Some advantages and disadvantages I see

Upsides
- More concise
- More readable
- Easier to refactor
- Better encapsulation (other modules do not need to know the internal file structure of other modules)

Downsides
- Namespace pollution
- Easier to run into circular imports (I guess the solution is to just use the full path for the import)
- Implicit dependencies

I think it can be beneficial for some types of modules, for instance pydantic models like this grouped in a single module. I think it is fine to use a mix and apply it wherever it benefits readability.

trailblazer/server/utils.py
```diff
@@ -1,14 +1,31 @@
...
 
-def parse_analysis_request(request: Request) -> AnalysisRequest:
+def parse_analyses_request(request: Request) -> AnalysesRequest:
     """Parse a request for retrieving analyses."""
     query_params = {}
     for key in request.args.keys():
         if key.endswith("[]"):
             query_params[key[:-2]] = request.args.getlist(key)
         else:
             query_params[key] = request.args.get(key)
```
ChrOertlin, line 14: is there a more robust way to identify keys from here? Maybe not for this PR.

trailblazer/dto/__init__.py
seallard: I'll leave it as is for now.

trailblazer/dto/analyses_response.py
seallard: No idea, it is a field on the database model for a job. Looking at the database, the values in the column seem to be case names.

seallard: It is not used in the analyses table, so I removed it from the response

trailblazer/dto/analyses_response.py
seallard: No, it is no longer used and can be removed

trailblazer/dto/__init__.py
henrikstranneheim: Rufff does not understand them:

```
trailblazer/dto/__init__.py:1:46: F401 [*] `trailblazer.dto.analyses_request.AnalysesRequest` imported but unused
trailblazer/dto/__init__.py:2:47: F401 [*] `trailblazer.dto.analyses_response.AnalysesResponse` imported but unused
trailblazer/dto/__init__.py:3:47: F401 [*] `trailblazer.dto.analysis_response.AnalysisResponse` imported but unused
trailblazer/dto/__init__.py:4:53: F401 [*] `trailblazer.dto.analysis_update_request.AnalysisUpdateRequest` imported but unused
trailblazer/dto/__init__.py:5:49: F401 [*] `trailblazer.dto.failed_jobs_request.FailedJobsRequest` imported but unused
trailblazer/dto/__init__.py:6:50: F401 [*] `trailblazer.dto.failed_jobs_response.FailedJobsResponse` imported but unused
trailblazer/services/__init__.py:1:51: F401 [*] `trailblazer.services.analysis_service.AnalysisService` imported but unused
trailblazer/services/__init__.py:2:46: F401 [*] `trailblazer.services.job_service.JobService` imported but unused
```

PR #346, opened 2024-01-09
Remove user from /analyses response

seallard:
## Description
The user associated with an analysis is no longer displayed in the analysis table in cigrid, so there is no need to return that data in the response. Removing it reduces the response time.

### Fixed
- Remove unnecessary data from response in /analyses endpoint


henrikstranneheim (approved):
👍

islean (approved):
🔥

PR #347, opened 2024-01-10
Add action for scanning images

seallard:
## Description
This PR adds an action which scans the trailblazer image for security issues and displays them in the security tab, for example [here](https://github.com/Clinical-Genomics/trailblazer/security/code-scanning?query=branch%3Ascan-containers+). The action is triggered on all pushes.

The action can also be configured to generate an SBOM if that is something we want as well.

### Added
- Action for scanning images for vulnerabilities


henrikstranneheim (approved):
Nice initiative 👍

PR #368, opened 2024-01-26
Update upload job status from slurm

seallard:
## Description
This PR adds logic to ensure that the status of ongoing upload jobs get updated. Part of the [implementation plan](https://github.com/Clinical-Genomics/cigrid-ui/issues/492#issuecomment-1905771598) in the [track analysis upload jobs story](https://github.com/Clinical-Genomics/cigrid-ui/issues/492).

### Added
- Dependency injection in web API endpoints
- Interface `SlurmService`
- Service `SlurmCLIService` (implements the `SlurmService` interface)
- New method on `JobService` called `update_upload_jobs`

## Notes
The `_update_analysis_from_tower_output` function overwrites the jobs stored on an analysis and will  disassociate the upload jobs for analyses managed by tower. This will be addressed in a separate PR.

Need to call the new logic (`update_upload_jobs`) from the `scan` cli command or similar.


trailblazer/services/slurm/slurm_service.py
```diff
@@ -0,0 +1,9 @@
+from abc import ABC, abstractmethod
+
+from trailblazer.constants import TrailblazerStatus
+
+
+class SlurmService(ABC):
```
seallard, line 6: Might be better to call this one `WorkflowManagerService` and create a service for tower as well implementing it.

trailblazer/store/crud/read.py
```diff
@@ -4,20 +4,26 @@
...
 
 
 class ReadHandler(BaseHandler):
     """Class for reading items in the database."""
 
+    def get_failed_jobs(self, since_when: datetime = None) -> list[dict[str, str | int]]:
+        """Return the number of failed jobs per category (name)."""
+        return self.get_nr_jobs_with_status_per_category(
+            status=TrailblazerStatus.FAILED, since_when=since_when
+        )
```
islean, line 25: I get confused by this method - why do we have a dict which goes from string to either string or int? Why not str -> str? Also, would prefer if it was called "get_number_of_failed_jobs" or "count_failed_jobs" or something similar. Here I would expect to get a "list[Job]" response.

islean (approved):
Looks nice! Just a bit confused - what is the intended use? Because right now the functionality is not put to use anywhere right? Might be nice with some small test of "update_upload_jobs". Also the new injection should be tested.

trailblazer/store/crud/read.py
seallard: Yes, the type hint is not helpful and the function name is confusing. The method was just moved around in this PR. I'll rename it

PR #373, opened 2024-01-29
Remove unused endpoints

seallard:
## Description
From what I can see, the `/update` and `/delete` endpoints are not used anywhere and can be removed. Please double check.

Neither of the endpoints show up in the logs on cg-prod-services. So I'm quite certain these can be removed.

### Fixed
- Remove endpoints which are not called


Vince-janv (approved):
Nice 🧹

PR #378, opened 2024-01-30
Update upload jobs in scan

seallard:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test
- [ ] activate stage: `us`
- [ ] install on stage:
    ```Shell
    ```
- [ ] install on stage:

### How to test
- [ ] login to ...
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] tests executed by

### DIFF
--- a/trailblazer/cli/core.py
+++ b/trailblazer/cli/core.py
@@ -6,14 +6,18 @@ from pathlib import Path
 import click
 import coloredlogs
 from sqlalchemy.orm import scoped_session
+from dependency_injector.wiring import inject, Provide
 
 import trailblazer
 from trailblazer.cli.utils.ls_helper import _get_ls_analysis_message
 from trailblazer.cli.utils.user_helper import is_existing_user, is_user_archived
 from trailblazer.constants import TRAILBLAZER_TIME_STAMP, FileFormat, TrailblazerStatus
+from trailblazer.containers import Container
 from trailblazer.environ import environ_email
 from trailblazer.io.controller import ReadFile
 from trailblazer.models import Config
+from trailblazer.services.analysis_service import AnalysisService
+from trailblazer.services.job_service import JobService
 from trailblazer.store.database import get_session, initialize_database
 from trailblazer.store.models import Analysis, User
 from trailblazer.store.store import Store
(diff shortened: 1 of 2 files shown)


trailblazer/cli/core.py
```diff
@@ -89,11 +95,14 @@ def init(context, reset, force):
 
 
 @base.command()
-@click.pass_context
-def scan(context):
+@inject
+def scan(
+    job_service: JobService = Provide[Container.job_service],
```
Vince-janv, line 100: Vert cool stuff. I'm not sure when this was merged but I'm lacking doc strings and type hints in `trailblazer/containers.py` and `trailblazer/server/wiring.py`. It's not super intuitive, and without some supporting information it risks becoming black magic only you can deal with 😅

trailblazer/clients/slurm_cli_client/utils.py
```diff
@@ -4,7 +4,7 @@
 
 def create_job_info_dto(job: SqueueJob) -> SlurmJobInfo:
     return SlurmJobInfo(
-        job_name=job.step,
+        name=job.step,
```
Vince-janv, line 7: yes 🙌

Vince-janv (approved):
Well done! It's going to be exciting to start adding upload jobs 🤞

trailblazer/cli/core.py
ChrOertlin: I don't thunk we should make ourselves dependent on this package without properly discussing it first. The dependency injector is not properly maintained any longer and we run the risk of having to re do this in the future. See here the initial developers stopped maintaining it and there is no one taking it over right now...
https://github.com/ets-labs/python-dependency-injector/pull/765 

Not sure if we should make ourselves vulnerable like this @henrikstranneheim, @islean , @diitaz93  what do you think?

Currently I think we solve dependency injection via de config.

seallard: Yeah, it will need to be replaced 👍

islean: @seallard @ChrOertlin I don't really see the benefit from merging this unapproved option if we know that we do not want it. I would have preferred it if we left out the injection until we have agreed on an option in situations like this.

seallard: The library was not introduced in this PR but I agree. Sometimes it is also nice to try things out, we can use trailblazer as a demo for whatever we settle on. There are 200 other repos to apply it in 😅

seallard: I don't think the solution we have (with the config) is working great. Usually when I comment on dependencies being instantiated in a CLI command or endpoint, it turns out it is impossible to do it in the config. So we need a better solution

islean: Oh, fair enough, if it was not introduced here you can disregard 😄 Did not intend for it to sound harsh, but if it was introduced here, it would probably have been nice to use a maintained package immediately. I agree that the config solution does have some issues though.

seallard: I really like the syntax of this package but it is not viable long term :/ While it has many users, there is another package that actually is maintained and is more widely used

henrikstranneheim: @seallard @islean @ChrOertlin Since the dependency is not new to TB, I see harm in using it. But, for the aforementioned 200 other apps it would be great with a discussion and well maintained packages before embarking on that odessey.

PR #382, opened 2024-02-01
Ensure upload jobs are not overwritten

seallard:
## Description
The current logic for updating analysis jobs overwrites all existing ones. Since we have added a separate category of jobs (upload jobs) which are associated with an analysis, this is not desirable.

Instead, only the *analysis* jobs are deleted by the `delete_analysis_jobs`.

### Fixed
- Ensure upload jobs are not deleted

### DIFF
--- a/trailblazer/store/crud/delete.py
+++ b/trailblazer/store/crud/delete.py
@@ -2,7 +2,7 @@ import logging
 
 from sqlalchemy.orm import Session
 
-from trailblazer.constants import TrailblazerStatus
+from trailblazer.constants import JobType, TrailblazerStatus
 from trailblazer.exc import MissingAnalysis, TrailblazerError
 from trailblazer.store.base import BaseHandler
 from trailblazer.store.database import get_session
@@ -18,7 +18,8 @@ class DeleteHandler(BaseHandler):
         """Delete all jobs linked to the given analysis."""
         session: Session = get_session()
         for job in analysis.jobs:
-            session.delete(job)
+            if job.job_type == JobType.ANALYSIS:
+                session.delete(job)
         session.commit()
 
     def delete_analysis(self, analysis_id: int, force: bool = False) -> None:
--- a/trailblazer/store/crud/update.py
+++ b/trailblazer/store/crud/update.py
@@ -26,10 +26,12 @@ LOG = logging.getLogger(__name__)
 class UpdateHandler(BaseHandler):
     """Class for updating items in the database."""
 
-    def update_analysis_jobs(self, analysis: Analysis, jobs: list[dict]) -> None:
+    def update_analysis_jobs(self, jobs: list[dict]) -> None:
         """Update jobs in the analysis."""
-        analysis.jobs = [Job(**job) for job in jobs]
         session: Session = get_session()
+        for job in jobs:
+            updated_job = Job(**job)
+            session.add(updated_job)
         session.commit()
 
     def update_user_is_archived(self, user: User, archive: bool = True) -> None:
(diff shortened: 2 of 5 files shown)


trailblazer/store/crud/update.py
```diff
@@ -26,10 +26,12 @@
...
     """Class for updating items in the database."""
 
-    def update_analysis_jobs(self, analysis: Analysis, jobs: list[dict]) -> None:
+    def update_analysis_jobs(self, jobs: list[dict]) -> None:
         """Update jobs in the analysis."""
-        analysis.jobs = [Job(**job) for job in jobs]
         session: Session = get_session()
+        for job in jobs:
+            updated_job = Job(**job)
+            session.add(updated_job)
```
islean, line 34: Will this also add them to analysis.jobs? Seems like we are changing the logic here by not setting that?

islean: And if they are added via the analysis.id we will just add them on top of the old ones, rather than switching?

trailblazer/store/crud/update.py
```diff
@@ -71,24 +73,22 @@ def update_analysis_jobs_from_slurm_jobs(
...
                 slurm_id=job.id,
                 name=job.step,
                 status=job.status,
                 started_at=job.started_at,
                 elapsed=job.time_elapsed,
             )
-            for job in squeue_result.jobs
-        ]
-        session: Session = get_session()
+            session.add(new_job)
```
islean, line 91: Same question here - why are we not replacing the old jobs?

trailblazer/store/crud/update.py
```diff
@@ -138,7 +138,7 @@ def _update_analysis_from_tower_output(
         analysis.logged_at = datetime.now()
         self.delete_analysis_jobs(analysis)
         jobs: list[dict] = tower_api.get_jobs(analysis.id)
-        self.update_analysis_jobs(analysis=analysis, jobs=jobs)
+        self.update_analysis_jobs(jobs)
```
islean, line 141: I think we probably need to keep the analysis as a parameter here or at least in some way be able to access its analysis.jobs property.

islean (commented):
Might be getting this confused but are we not changing more behaviour than we intend here?

trailblazer/store/crud/update.py
seallard: Yes, the job dictionary contains the analysis id. And no, they are deleted before this function is called. Would probably make sense to delete them in here instead, but I'm minimizing the changes for now

trailblazer/store/crud/update.py
seallard: We are, all the analysis jobs are deleted above. The old logic deleted all jobs and then (unnecessarily) also overwrote an empty list with `analysis.jobs = ...`.

islean (approved):
Fair enough then, seems as if there was some cryptic logic from before.

trailblazer/store/crud/update.py
```diff
@@ -26,10 +26,12 @@
 class UpdateHandler(BaseHandler):
     """Class for updating items in the database."""
 
-    def update_analysis_jobs(self, analysis: Analysis, jobs: list[dict]) -> None:
+    def update_analysis_jobs(self, jobs: list[dict]) -> None:
         """Update jobs in the analysis."""
```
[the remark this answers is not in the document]
seallard: It will always be called in a context with analysis jobs, but I think the name you suggest works 👍

trailblazer/store/crud/delete.py
```diff
@@ -18,7 +18,8 @@ def delete_analysis_jobs(self, analysis: Analysis) -> None:
         """Delete all jobs linked to the given analysis."""
         session: Session = get_session()
         for job in analysis.jobs:
-            session.delete(job)
+            if job.job_type == JobType.ANALYSIS:
+                session.delete(job)
```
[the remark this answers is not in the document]
seallard: It is rather ensuring the old behavior, but I get what you mean. I'll add a test to to prevent any bugs being introduced in the future that could cause the upload jobs to be deleted again

PR #383, opened 2024-02-02
Refactor endpoint for creating analysis

seallard:
## Description
Refactor `add-pending-analysis` endpoint to follow the pattern of
request -> pydantic model -> service layer -> database layer -> database model -> service layer -> pydantic model -> response.

### Fixed
- Refactor endpoint for creating analyses
- Add new integration test

### DIFF
--- a/trailblazer/dto/create_analysis_request.py
+++ b/trailblazer/dto/create_analysis_request.py
@@ -1,0 +1,15 @@
+from pydantic import BaseModel
+
+from trailblazer.constants import JobType, TrailblazerPriority, TrailblazerTypes, WorkflowManager
+
+
+class CreateAnalysisRequest(BaseModel):
+    case_id: str
+    email: str | None = None
+    type: JobType | None = JobType.ANALYSIS
+    config_path: str
+    out_dir: str
+    priority: TrailblazerPriority
+    data_analysis: TrailblazerTypes
+    ticket_id: str | None = None
+    workflow_manager: WorkflowManager | None = None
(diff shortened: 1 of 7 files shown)


trailblazer/dto/create_analysis_request.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CreateAnalysisRequest(BaseModel):
+    case_id: str
+    email: str | None = None
+    config_path: str
+    out_dir: str
+    priority: TrailblazerPriority
+    data_analysis: TrailblazerTypes
+    ticket_id: str | None = None
+    workflow_manager: WorkflowManager | None = None
```
seallard, line 14: Did I miss any restrictions here? Maybe some of the fields should not be allowed to be None?

trailblazer/dto/create_analysis_request.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CreateAnalysisRequest(BaseModel):
+    case_id: str
+    email: str | None = None
+    config_path: str
+    out_dir: str
+    priority: TrailblazerPriority
+    data_analysis: TrailblazerTypes | None = None
+    ticket_id: str | None = None
+    workflow_manager: WorkflowManager | None = None
```
seallard, line 14: Did I miss any restrictions here? Maybe some of the fields should not be allowed to be None? The database columns are nullable for those, but might not be what we really want. The old method had these as nullable
```python
     email: str = None
     data_analysis: str = None
     ticket_id: str = None
     workflow_manager: str = None
```
To avoid breaking changes, we should probably stick to those

henrikstranneheim: I think email can be none, none of the others.

henrikstranneheim (approved):
Clean 👍 !

trailblazer/dto/create_analysis_request.py
seallard: I'll leave it as is and we can refactor it in a separate PR

PR #385, opened 2024-02-02
Fix circular import

seallard:


trailblazer/server/wiring.py
```diff
@@ -1,10 +1,10 @@
 from trailblazer.containers import Container
-import trailblazer.server.api
-import trailblazer.cli.core
 
 container = Container()
 
 
 def setup_dependency_injection() -> Container:
+    import trailblazer.server.api
+    import trailblazer.cli.core
```
islean, line 8: ooo that is cool!

islean (approved):
Nice! Do the test run in isolation now?

seallard:
> Nice! Do the test run in isolation now?

Yes! Somewhat ugly fix, but we will swap out this component anyway

PR #386, opened 2024-02-06
Add filter parameters to analyses endpoint

seallard:
## Description
Additional filter parameter on the `/analyses` endpoint required by https://github.com/Clinical-Genomics/cg/pull/2895.

### Added
- Filtering analyses on case id

### DIFF
--- a/trailblazer/store/base.py
+++ b/trailblazer/store/base.py
@@ -63,8 +63,6 @@ class BaseHandler:
             filters.append(AnalysisFilter.FILTER_BY_TYPES)
         if query.case_id:
             filters.append(AnalysisFilter.FILTER_BY_CASE_ID)
-        if query.most_recent:
-            filters.append(AnalysisFilter.FILTER_BY_MOST_RECENT)
 
         return apply_analysis_filter(
             filter_functions=filters,
--- a/trailblazer/store/base.py
+++ b/trailblazer/store/base.py
@@ -61,6 +61,10 @@ class BaseHandler:
             filters.append(AnalysisFilter.FILTER_BY_STATUSES)
         if query.type:
             filters.append(AnalysisFilter.FILTER_BY_TYPES)
+        if query.case_id:
+            filters.append(AnalysisFilter.FILTER_BY_CASE_ID)
+        if query.most_recent:
+            filters.append(AnalysisFilter.FILTER_BY_MOST_RECENT)
 
         return apply_analysis_filter(
             filter_functions=filters,
@@ -70,6 +74,7 @@ class BaseHandler:
             priorities=query.priority,
             statuses=query.status,
             types=query.type,
+            case_id=query.case_id,
         )
 
     def get_visible_analyses(self, analyses: Query) -> Query:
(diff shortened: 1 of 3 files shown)


trailblazer/store/filters/analyses_filters.py
```diff
@@ -80,6 +80,10 @@ def filter_analyses_by_empty_comment(analyses: Query, **kwargs) -> Query:
     return analyses.filter(sqlalchemy.or_(Analysis.comment.is_(None), Analysis.comment == ""))
 
 
+def filter_analyses_by_most_recent(analyses: Query, **kwargs) -> Query:
+    return analyses.order_by(Analysis.started_at.desc()).limit(1)
```
islean, line 84: Feels slightly odd to call analyses and only want a single analysis back. Maybe this should be a different endpoint for cases? Might be too much of a rewrite though.

islean (approved):
It will most likely do what you want. Just feel a bit iffy about the most_recent parameter.

trailblazer/store/filters/analyses_filters.py
islean: And would it ever be set in isolation?

trailblazer/store/base.py
```diff
@@ -61,6 +61,10 @@ def get_filtered_analyses(self, analyses: Query, query: AnalysesRequest) -> Quer
             filters.append(AnalysisFilter.FILTER_BY_STATUSES)
         if query.type:
             filters.append(AnalysisFilter.FILTER_BY_TYPES)
+        if query.case_id:
+            filters.append(AnalysisFilter.FILTER_BY_CASE_ID)
+        if query.most_recent:
+            filters.append(AnalysisFilter.FILTER_BY_MOST_RECENT)
```
ChrOertlin, line 67: can we break this function up instead of adding to this? It seems like this can end up with an amount of if statements proportionate to the number of attributes an analyses can be filtered on. Or at least, let's not keep adding to this.

How do we now at any given point in a function call what is actually being filtered on when we apply this function without needing to check what the `query` actually has stored.

Also, `analyeses: Query` and `query: AnalysesRequest` is extremely confusing

trailblazer/store/filters/analyses_filters.py
ChrOertlin, line 84: nice!

trailblazer/store/base.py
islean: Could we not rewrite these methods where we have a request DTO in some way like this:

```
filters = []
for field in query.fields():
   if query.field:
       filters.append(SOME ENUM MAGIC FOR CONVERTING)
apply_filters
```

trailblazer/store/filters/analyses_filters.py
seallard: True, it was not really necessary - the default sort field and order is enough. Removed it

trailblazer/store/base.py
seallard: I think we can consider some alternative pattern too. I'd like something like a `analyses_request_filters` constant and move the null checks into the filters

seallard: Leaving this for future PR:s

PR #398, opened 2024-02-21
Add summary endpoint

seallard:
## Description
Add endpoint for retrieving summaries for analyses in orders. Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/42.

### Added
- Endpoint for retrieving summaries for analyses in orders


trailblazer/server/utils.py
```diff
@@ -39,3 +40,12 @@ def stringify_timestamps(data: dict) -> dict[str, str]:
...
 def parse_job_create_request(request: Request) -> CreateJobRequest:
     return CreateJobRequest.model_validate(request.json)
+
+
+def parse_summaries_request(request: Request) -> SummariesRequest:
+    query_params = {
+        key[:-2]: request.args.getlist(key)
+        for key in request.args.keys()
+        if key_has_list_of_values(key)
+    }
```
islean, line 50: I find it slightly confusing with the indexes, but maybe it just needs getting used to. What would be the difference from doing a model_validate on the request.args?

trailblazer/services/analysis_service/utils.py
```diff
@@ -0,0 +1,41 @@
...
+def get_status_count(analyses: list[Analysis], status: TrailblazerStatus) -> int:
+    return len([a for a in analyses if a.status == status])
+
+
+def create_summary(analyses: list[Analysis]) -> Summary:
+    total: int = len(analyses)
+    delivered: int = get_status_count(analyses=analyses, status=TrailblazerStatus.COMPLETED)
+    running: int = get_status_count(analyses=analyses, status=TrailblazerStatus.RUNNING)
+    cancelled: int = get_status_count(analyses=analyses, status=TrailblazerStatus.CANCELLED)
+    failed = get_status_count(analyses=analyses, status=TrailblazerStatus.FAILED)
```
islean, line 17: This will work! It might be cool/concise to use `count([a.status for a in analyses])` or whatever the syntax is. Up to you though, this might just be clearer.

trailblazer/services/analysis_service/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    return Summary(
+        order_id=analyses[0].order_id,
```
islean, line 19: Very minor, but you could consider passing the order_id as only input to this method and fetch analyses inside it. Then you would avoid order_id->analyses->order_id. Feel free to disregard though.

trailblazer/services/analysis_service/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    analysis_data["user"] = analysis.user.to_dict() if analysis.user else None
+    return AnalysisResponse.model_validate(analysis_data)
+
+
+def create_update_analyses_response(analyses: list[Analysis]) -> UpdateAnalysesResponse:
+    response_data: list[dict] = []
+    for analysis in analyses:
+        analysis_data = analysis.to_dict()
+        response_data.append(analysis_data)
+    return UpdateAnalysesResponse(analyses=response_data)
```
islean, line 41: These were just moved here right?

islean (approved):
Looks great!

trailblazer/server/api.py
```diff
@@ -131,6 +134,17 @@ def update_analysis(
...
 
+@blueprint.route("/summary", methods=["GET"])
+@inject
+def get_summaries(analysis_service: AnalysisService = Provide[Container.analysis_service]):
+    try:
+        request_data: SummariesRequest = parse_summaries_request(request)
+        response: SummariesResponse = analysis_service.get_summaries(request_data)
+        return jsonify(response.model_dump()), HTTPStatus.OK
+    except ValidationError as error:
+        return jsonify(error=str(error)), HTTPStatus.BAD_REQUEST
```
islean, line 145: Would we want to add some validation that we actually have some analyses/order_ids in place?

seallard: I think it is fine to get an empty list as a response. The caller does not request a specific resource and should receive an error if there are none

seallard: If the caller does not provide any order id:s I think it is fine that we return an empty successful response.
I also don't think it is an error that there are no analyses available for the provided orders

trailblazer/services/analysis_service/utils.py
seallard: Yes!

trailblazer/services/analysis_service/utils.py
seallard: I'd like that, but seems like `count` only supports direct checks on the elements in the list - not attributes on the object

trailblazer/server/utils.py
seallard: The `key[:-2]` discards the leading `[]` before the name of the query parameter. It is a standard to add a leading `[]` before a list query param. I moved it into the pydantic model instead

trailblazer/services/analysis_service/utils.py
seallard: Passed the order id, will keep the fetching of the orders outside though

PR #402, opened 2024-03-06
Add /auth endpoint

seallard:
## Description
Add `/auth` endpoint. A step in supporting the authorization code flow, see [docs](https://developers.google.com/identity/protocols/oauth2/web-server#exchange-authorization-code). The documentation describes the entire flow, this PR implements Step 5: `Exchange authorization code for refresh and access tokens`.

This endpoint will be used when a user signs in via Google to exchange the resulting authorization code for a refresh and access token. We can likely extract this service in the future, added to backlog here https://github.com/Clinical-Genomics/cg/issues/3027.

Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/58. The goal is to ensure users in Cigrid do not have to re-login during the workday.

The new components:
- Authentication service: contains logic for authentication
- Encryption service: contains logic for encrypting data
- Google OAuth client: web API client for Googles OAuth endpoints
- Google API client: web API client for Googles API
- New column `refresh_token` in user table

The authentication service does the following
1. Exchange the authorization code for an access token and refresh token via the Google OAuth client
2. Use the access token to retrieve the users email via the Google API client
3. Retrieve the user with matching email from the database if it exists
4. Encrypt the refresh token and store it on the user
5. Return the access token

This logic is exposed via the `/auth` endpoint.

### Added
- `/auth` endoint


trailblazer/services/encryption_service/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    return plain_text.decode()
+
+
+def _encrypt_with_aes(data: bytes, secret_key: str) -> bytes:
+    secret_key = secret_key.encode()
+    initialization_vector: bytes = os.urandom(BLOCK_SIZE)
+
+    cipher = Cipher(
+        algorithms.AES(secret_key),
+        modes.CBC(initialization_vector),
```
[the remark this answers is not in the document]
seallard: Switched to GCM

trailblazer/services/encryption_service/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    return initialization_vector + cipher_text
+
+
+def _decrypt_with_aes(data: bytes, secret_key: str) -> bytes:
+    secret_key: bytes = secret_key.encode()
+    initialization_vector: bytes = data[:BLOCK_SIZE]
+
+    cipher = Cipher(
+        algorithms.AES(secret_key),
+        modes.CBC(initialization_vector),
```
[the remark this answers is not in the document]
seallard: Switched to GCM

trailblazer/server/api.py
```diff
@@ -63,6 +68,19 @@ def before_request():
         return abort(403, f"{user_data['email']} doesn't have access")
 
 
+@blueprint.route("/auth", methods=["POST"])
+@inject
+def authenticate(auth_service: AuthenticationService = Provide[Container.auth_service]):
+    try:
+        request_data = CodeExchangeRequest.model_validate(request.json)
```
ChrOertlin, line 75: This is a bit unclear to me, where does the `request.json` come from?

ChrOertlin (approved):
Looks good!

trailblazer/server/api.py
seallard: Me too! Read a little about the Flask library and apparently it makes the `request` object available in functions decorated with `@app.route` or `@blueprint.route`. It represents the incoming web request https://flask.palletsprojects.com/en/3.0.x/api/#flask.request.

I can't find any way of explicitly passing it, even if it could be nice to do so

ChrOertlin: Ok, good to know. Thanks!

trailblazer/services/encryption_service/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+def decrypt_data(data: str, secret_key: str) -> str:
+    data: bytes = base64.b64decode(data)
+    plain_text: bytes = _decrypt_with_aes(data=data, secret_key=secret_key)
+    return plain_text.decode()
+
+
+def _encrypt_with_aes(data: bytes, secret_key: str) -> bytes:
+    cipher = AESGCM(base64.b64decode(secret_key))
+    nonce: bytes = os.urandom(NONCE_SIZE)
```
islean, line 22: A bit uncertain how the os.urandom works, but does it give a random byte string _up to_ 12 in length or exactly 12? If the former I think the decryption will fail since the length can vary

trailblazer/store/models.py
```diff
@@ -40,6 +40,7 @@ class User(Model):
     id = Column(types.Integer, primary_key=True)
     is_archived = Column(types.Boolean, default=False)
     name = Column(types.String(128))
+    refresh_token = Column(types.String(128))
```
islean, line 43: similar question here - do we know how long the token can be?

tests/integration/services/test_authentication_service.py
```diff
@@ -0,0 +1,18 @@
...
+
+from trailblazer.services.authentication_service.authentication_service import AuthenticationService
+from trailblazer.store.models import User
+
+
+def test_exchange_code(authentication_service: AuthenticationService, user_email: str):
+    # GIVEN an authentication service
+
+    # WHEN exchanging the authorization code
+    token: str = authentication_service.authenticate("auth_code")
```
islean, line 11: Nice test! It looks a bit finicky to try much in a unit test though (I think this is as far as it goes perhaps). We should probably test this quite extensively so that the tokens actually work as intended as well.

trailblazer/clients/authentication_client/google_oauth_client.py
```diff
@@ -0,0 +1,31 @@
...
+        """Exchange the authorization code for an access token and refresh token."""
+        request = GetTokensRequest(
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            code=authorization_code,
+            redirect_uri=self.redirect_uri,
+        )
+        data: str = request.model_dump_json()
+
+        response = requests.post(self.token_uri, data=data)
```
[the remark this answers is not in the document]
seallard: You are right, changed to `model_dump`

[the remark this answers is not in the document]
islean: It might be better with `model_dump_json` though, I think it supports serializing date time objects for example. But then you would need to set it as the json parameter in the requests.post. I.e. `requests.post(self.token_uri, json=json_dump)`. Might need more work though, haven't used it too much.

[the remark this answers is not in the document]
seallard: Google uses `data=...` in their example, so I'll stick to it for now!

trailblazer/server/api.py
```diff
@@ -47,6 +50,8 @@
 @blueprint.before_request
 def before_request():
     """Authentication that is run before processing requests to the application"""
+    if request.endpoint == "api.authenticate":
+        return
```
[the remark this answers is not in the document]
seallard: Yes

trailblazer/services/encryption_service/utils.py
seallard: Exactly 12! So it won't be an issue (this module is covered by tests as well)

trailblazer/store/models.py
seallard: Changed to text

seallard: https://github.com/Clinical-Genomics/trailblazer/pull/402/commits/c701b25fa1ec3ccf464b4ab492bd9fcf815a5642

trailblazer/clients/authentication_client/google_oauth_client.py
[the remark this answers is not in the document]
seallard: https://github.com/Clinical-Genomics/trailblazer/pull/402/commits/8f577c23e578b7ac434d949795718a0d6808a4ba

tests/integration/services/test_authentication_service.py
seallard: I'll have to revise parts of this logic for sure once I start implementing the frontend components

PR #404, opened 2024-03-08
Refresh token

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/59.
This PR adds a new endpoint `/auth/refresh` which enables authenticated users to refresh their access token.
It calls the Google OAuth endpoint documented [here](https://developers.google.com/identity/protocols/oauth2/web-server#offline).

### Added
- `/auth/refresh`


trailblazer/server/api.py
```diff
@@ -82,6 +82,18 @@ def authenticate(auth_service: AuthenticationService = Provide[Container.auth_se
...
 
+@blueprint.route("/auth/refresh", methods=["GET"])
+@inject
+def refresh_token(auth_service: AuthenticationService = Provide[Container.auth_service]):
+    """Refresh access token."""
+    user: User = g.current_user
+    try:
+        token: str = auth_service.refresh_access_token(user.id)
+        return jsonify({"access_token": token}), HTTPStatus.OK
+    except AuthenticationError:
```
ChrOertlin, line 93: `refresh_access_token` also throws a `UserNotFoundError` should we catch this here too?

ChrOertlin (approved):
Looks good

trailblazer/server/api.py
seallard: Yes! When I look at the code now, I realize that the user will always exist since that is validated in the `before_request`. I removed the raised `UserNotFoundError`

seallard: Here is the `before_request` function in `api.py` which always runs before a request is processed
```python
@blueprint.before_request
def before_request():
    """Authentication that is run before processing requests to the application"""
    ...
    if auth_header := request.headers.get("Authorization"):
        jwt_token = auth_header.split("Bearer ")[-1]
    else:
        return abort(403, "no JWT token found on request")

    user_data: Mapping = jwt.decode(jwt_token, verify=False)
    if user := store.get_user(email=user_data["email"], exclude_archived=True):
        g.current_user = user
    else:
        return abort(403, f"{user_data['email']} doesn't have access")
```

So we will abort the request immediately if the user does not exist in this instance.

ChrOertlin: Nice, then we don't need to check for user existence elswehre indeed!

PR #406, opened 2024-03-11
Improve env var name

seallard:
An env var was vaguely named.


trailblazer/containers.py
```diff
@@ -18,7 +18,7 @@ class Container(containers.DeclarativeContainer):
     oauth_client_secret: str = os.environ.get("GOOGLE_CLIENT_SECRET")
     oauth_redirect_uri: str = os.environ.get("GOOGLE_REDIRECT_URI")
     google_oauth_base_url: str = os.environ.get("GOOGLE_OAUTH_BASE_URL")
-    encryption_key: str = os.environ.get("SECRET_KEY")
+    encryption_key: str = os.environ.get("ENCRYPTION_KEY")
```
islean, line 21: I don't think I like "encryption_key". In many cryptographic schemas there are both public and private/secret keys. So I think this is probably a bit more unclear to me.

seallard: Ah, what do you think it should be called?

islean: maybe AES_PRIVATE_KEY might be best?

PR #410, opened 2024-03-13
Return id token

seallard:
We want to return the `id_token` instead of the access token. The access token is only necessary if we want to access any information about the user via a Google API.

Since we only use the token for authenticating the user, the access token is not needed. We want the token which serves as proof of the users identity.

Adjusts https://github.com/Clinical-Genomics/streamline-delivery/issues/58.


ChrOertlin (approved):
Looks good.

PR #411, opened 2024-03-14
Return id token when refreshing

seallard:
We want to return the id token in the refresh endpoint as well. Adjusts https://github.com/Clinical-Genomics/streamline-delivery/issues/59.

### DIFF
--- a/trailblazer/server/api.py
+++ b/trailblazer/server/api.py
@@ -88,8 +88,8 @@ def refresh_token(auth_service: AuthenticationService = Provide[Container.auth_s
     """Refresh access token."""
     user: User = g.current_user
     try:
-        token: str = auth_service.refresh_access_token(user.id)
-        return jsonify({"access_token": token}), HTTPStatus.OK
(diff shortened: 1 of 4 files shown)


islean (approved):
Looks good!

trailblazer/server/api.py
```diff
@@ -88,8 +88,8 @@ def refresh_token(auth_service: AuthenticationService = Provide[Container.auth_s
     """Refresh access token."""
     user: User = g.current_user
     try:
-        token: str = auth_service.refresh_access_token(user.id)
-        return jsonify({"access_token": token}), HTTPStatus.OK
+        token: str = auth_service.refresh_token(user.id)
+        return jsonify({"token": token}), HTTPStatus.OK
```
islean, line 92: One though- you write that we want to return the id token "as well" but this only returns that token right?

seallard: Ah, sorry that was unclear! I meant "as well" as in we want to return the id token both from `/auth` and `/auth/refresh`

PR #413, opened 2024-03-15
Support filtering delivered analyses

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/74.


trailblazer/store/filters/analyses_filters.py
```diff
@@ -80,6 +80,13 @@ def filter_analyses_by_empty_comment(analyses: Query, **kwargs) -> Query:
     return analyses.filter(sqlalchemy.or_(Analysis.comment.is_(None), Analysis.comment == ""))
 
 
+def filter_analyses_by_delivered(analyses: Query, delivered: bool | None, **kwargs) -> Query:
+    if delivered == True:
+        return analyses.filter(Analysis.delivery != None)
+    if delivered == False:
+        return analyses.filter(Analysis.delivery is None)
```
islean, line 87: Just to double check - does this comparison work? I feel as if I always have to do "=="-comparisons in these filters even though pycharm complains.

trailblazer/store/base.py
```diff
@@ -63,6 +63,8 @@ def get_filtered_analyses(self, analyses: Query, query: AnalysesRequest) -> Quer
             filters.append(AnalysisFilter.BY_TYPES)
         if query.case_id:
             filters.append(AnalysisFilter.BY_CASE_ID)
+        if query.delivered is not None:
```
islean, line 66: Nice catch with "is not None" 💪

islean (approved):
Looks good to me! Might want to add a test for it though?

trailblazer/store/filters/analyses_filters.py
seallard: No, you are right - this was not correct. Fixed here https://github.com/Clinical-Genomics/trailblazer/pull/413/commits/3763426b539c9363ee851c2c9eb74a37dde83824

seallard:
> Looks good to me! Might want to add a test for it though?

I'll leave it as is, we can catch it in an integration test for the new deliver feature

PR #414, opened 2024-03-15
Refactor analyses filtering

seallard:
## Description
This PR refactors some of the analyses filtering logic to better adhere to the existing design pattern we have established.

- Extract filters for pagination, sorting, workflow.
- Validate sort order and sort field values


tests/integration/endpoints/test_get_analyses.py
```diff
@@ -98,7 +98,7 @@ def test_get_analyses_without_comments(client: FlaskClient, analyses: list[Analy
     # GIVEN analyses with and without comments
 
     # WHEN retrieving all analyses without comments
-    response = client.get(f"/api/v1/analyses?pageSize={len(analyses)}&comment[]=")
+    response = client.get(f"/api/v1/analyses?pageSize={len(analyses)}&hasComment=False")
```
seallard, line 101: Adjust cigrid accordingly

trailblazer/store/filters/analyses_filters.py
```diff
@@ -111,6 +149,13 @@ def apply_analysis_filter(
     status: str | None = None,
     statuses: list[str] | None = None,
     types: list[str] | None = None,
+    workflow: str | None = None,
+    has_comment: bool | None = None,
+    is_visible: bool | None = None,
+    page: int = 1,
```
seallard, line 155: Not sure about this, might be cleaner to just keep the nullable pattern?

trailblazer/store/crud/read.py
```diff
@@ -203,3 +204,46 @@ def get_analyses_by_order_id(self, order_id: int) -> list[Analysis]:
             filter_functions=[AnalysisFilter.BY_ORDER_ID],
             order_id=order_id,
         ).all()
+
+    def get_analyses(self, request: AnalysesRequest) -> tuple[list[Analysis], int]:
```
ChrOertlin, line 208:
```suggestion
    def get_analyses_to_paginate(self, request: AnalysesRequest) -> tuple[list[Analysis], int]:
```
Since it is in the CRUD. A function name that outlines the intent of the function more clearly is beneficial here.

ChrOertlin (approved):
Looks good.

trailblazer/store/crud/read.py
seallard: I like it 👍 What do you think of `get_analyses_by_request` as an option?

seallard: Nvm, I prefer `get_analyses_to_paginate`

tests/integration/endpoints/test_get_analyses.py
seallard: https://github.com/Clinical-Genomics/cigrid-ui/pull/550

PR #415, opened 2024-03-19
Patch analyses search

seallard:
Only hidden analyses were displayed when searching.


islean:
Tested in stage and search now shows both visible and hidden analyses! Also, hidden are filtered out if search is not specified.

PR #417, opened 2024-03-20
(Streamline delivery) Use latest analysis per case

islean:
## Description

When counting analysis statuses within an order, we should only use the latest analysis within each case. This PR adds a filter which makes sure that only the analysis with the latest `started_at` per case is used.

### Added

- Filter by latest_per_case

### Fixed

- Order summaries only use the latest analysis in each case.

### How to prepare for test
- [ ] activate stage: `us`
- [ ] install on stage:
    ```Shell
    ```
- [ ] install on stage:

### How to test
- [ ] login to ...
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] tests executed by

### DIFF
--- a/trailblazer/store/crud/read.py
+++ b/trailblazer/store/crud/read.py
@@ -195,7 +195,7 @@ class ReadHandler(BaseHandler):
             job_id=job_id,
         ).first()
 
-    def get_latest_analyses_in_an_order(self, order_id: int) -> list[Analysis]:
+    def get_latest_analyses_for_order(self, order_id: int) -> list[Analysis]:
         """Returns the latest analysis per case in the given order."""
         return apply_analysis_filter(
             analyses=self.get_query(Analysis),
--- a/trailblazer/store/crud/read.py
+++ b/trailblazer/store/crud/read.py
@@ -7,10 +7,7 @@ from sqlalchemy.orm import Query
 from trailblazer.constants import JobType, TrailblazerStatus
 from trailblazer.dto.analyses_request import AnalysesRequest
 from trailblazer.store.base import BaseHandler
-from trailblazer.store.filters.analyses_filters import (
-    AnalysisFilter,
-    apply_analysis_filter,
-)
+from trailblazer.store.filters.analyses_filters import AnalysisFilter, apply_analysis_filter
 from trailblazer.store.filters.job_filters import JobFilter, apply_job_filters
 from trailblazer.store.filters.user_filters import UserFilter, apply_user_filter
 from trailblazer.store.models import Analysis, Job, User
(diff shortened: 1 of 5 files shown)


islean:
Hmmm, seems not to work currently:

image

seallard:
If it works it works! But I'd prefer a different structure of the logic to make it easier to understand and modify (if possible). Might be worth making it less general/efficient to gain some readability

islean:
> If it works it works! But I'd prefer a different structure of the logic to make it easier to understand and modify (if possible). Might be worth making it less general/efficient to gain some readability

Do you prefer fetching all analyses for the order_id and then handling it in Python?

islean:
Would ideally want some more feedback on whether the design is good or bad here

trailblazer/store/filters/analyses_filters.py
```diff
@@ -107,6 +109,43 @@ def filter_analyses_by_workflow(analyses: Query, workflow: Workflow, **kwargs) -
...
+        FROM
+            Analysis
+        JOIN
+            (SELECT
+                case_id, MAX(started_at) started_at
+            FROM
+                analyses
+            GROUP BY
+                case_id) latest_started_analyses
+        USING (case_id, started_at)
```
ChrOertlin, line 127: I really like this for such complex queries. Helps a lot in understanding!

trailblazer/store/crud/read.py
```diff
@@ -198,10 +195,11 @@ def get_job_by_id(self, job_id: int) -> Job | None:
             job_id=job_id,
         ).first()
 
-    def get_analyses_by_order_id(self, order_id: int) -> list[Analysis]:
+    def get_latest_analyses_in_an_order(self, order_id: int) -> list[Analysis]:
```
ChrOertlin, line 198:
```suggestion
    def get_latest_analyses_in_order(self, order_id: int) -> list[Analysis]:
```
the order is specified with id and not random, flavour

ChrOertlin (approved):
Looks good!

trailblazer/store/crud/read.py
islean: Hmm yeah, true. Although `get_latest_analyses_in_order` makes it sound as if it is an ordered list rather than them belonging to the same order... @ChrOertlin

islean:
Seems to function as expected now. Most mismatches come from RSYNC analyses also having the same order_id so we need to get rid of those too. But that is out of scope.

trailblazer/store/crud/read.py
henrikstranneheim: How about "from_order"?

seallard:
> > If it works it works! But I'd prefer a different structure of the logic to make it easier to understand and modify (if possible). Might be worth making it less general/efficient to gain some readability
> 
> Do you prefer fetching all analyses for the order_id and then handling it in Python? Given that the number of analyses in an order won't be too many, that would probably be feasible, but then we would need to fetch all analyses in the order, sort them on case and then loop through each case-analyses bundle and keep only one from each.

I'm not sure! Maybe we can measure how much faster this is and make a judgement call accordingly? But I'm fine with this to be honest 👍

islean:
> > > If it works it works! But I'd prefer a different structure of the logic to make it easier to understand and modify (if possible). Might be worth making it less general/efficient to gain some readability
> > 
> > 
> > Do you prefer fetching all analyses for the order_id and then handling it in Python? Given that the number of analyses in an order won't be too many, that would probably be feasible, but then we would need to fetch all analyses in the order, sort them on case and then loop through each case-analyses bundle and keep only one from each.
> 
> I'm not sure! Maybe we can measure how much faster this is and make a judgement call accordingly? But I'm fine with this to be honest 👍

Oh, in this case I do not think it is a matter of performance. The bottleneck seem to be cg either way and looping through a list of at most 50 cases twice should be very fast. I just feel that that option is clunkier. And if we are hoping to move more logic to in other instances I do think we will see more of this.

PR #422, opened 2024-03-27
(Streamline delivery) Filter out RSYNC from summaries

islean:
## Description

When counting analyses in an order, the RSYNC analyses should be excluded. This PR removes them, although we would like to deprecate them being an analysis object altogether in the future.

### Fixed

- RSYNC analyses are excluded in the summary.

### How to prepare for test
- [ ] activate stage: `us`
- [ ] install on stage:
    ```Shell
    ```
- [ ] install on stage:

### How to test
- [ ] login to ...
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] tests executed by

### DIFF
--- a/trailblazer/services/analysis_service/analysis_service.py
+++ b/trailblazer/services/analysis_service/analysis_service.py
@@ -74,7 +74,9 @@ class AnalysisService:
         summaries: list[Summary] = []
         for order_id in request_data.order_ids:
             analyses: list[Analysis] = self.store.get_latest_analyses_for_order(order_id)
-            analyses_without_rsync: list[Analysis] = [analysis for analysis in analyses if analysis.workflow != Workflow.RSYNC]
-            summary: Summary = create_summary(analyses=analyses, order_id=order_id)
+            analyses_without_rsync: list[Analysis] = [
+                analysis for analysis in analyses if analysis.workflow != Workflow.RSYNC
+            ]
+            summary: Summary = create_summary(analyses=analyses_without_rsync, order_id=order_id)
             summaries.append(summary)
         return SummariesResponse(summaries=summaries)


trailblazer/constants.py
```diff
@@ -41,6 +41,7 @@ class Workflow(StrEnum):
     MIP_DNA: str = "MIP-DNA"
     MIP_RNA: str = "MIP-RNA"
     MUTANT: str = "MUTANT"
+    RSYNC: str = "RSYNC"
```
islean, line 44: Should I refactor this to the "proper" ENUM handling in this PR @Vince-janv ?

seallard: I say do it in another PR!

trailblazer/services/analysis_service/analysis_service.py
```diff
@@ -73,6 +74,9 @@ def get_summaries(self, request_data: SummariesRequest) -> SummariesResponse:
         summaries: list[Summary] = []
         for order_id in request_data.order_ids:
             analyses: list[Analysis] = self.store.get_latest_analyses_for_order(order_id)
-            summary: Summary = create_summary(analyses=analyses, order_id=order_id)
+            analyses_without_rsync: list[Analysis] = [
+                analysis for analysis in analyses if analysis.workflow != Workflow.RSYNC
```
seallard, line 78: This filtering might be done faster in the database query instead, but lets check if this adds a noticeable delay first

seallard (approved):
👍 
We should consider moving the filtering into the database query since this endpoint might be somewhat of a bottle neck for cigrid users

trailblazer/services/analysis_service/analysis_service.py
islean: but I think a filter for "not rsync" would become obsolete in the near future.

trailblazer/constants.py
Vince-janv: I agree with SA. I think these  ENUM refactorings can be quite error-prone so doing them in separate PRs would be good IMO

trailblazer/services/analysis_service/analysis_service.py
seallard: Yeah, true. When we retrieve orders from cg, we do call this endpoint once per order which could motivate speeding it up a bit. But I agree, lets leave it

islean: Resolved in private chat to leave as is for now.

PR #423, opened 2024-03-27
Refactor logic for updating analyses

seallard:
## Description
Refactor logic for updating analyses with information from slurm. The logic does not belong in the data access layer.

New design
- Analysis service: uses a job service to update the analysis (status, progress etc.).
- Job service: uses a slurm service to retrieve information about jobs from slurm.
- Slurm service: defines an interface for retrieving information about jobs from slurm.
- Slurm CLI service: implements the slurm service interface, uses a Slurm CLI client.
- Slurm CLI client: retrieves information from slurm via the CLI (wraps the old logic).

I'll remove all the deprecated logic in a separate PR.

### Fixed
- Extract logic for parsing slurm from data access layer
- Abstract away slurm interactions to enable going via API or CLI

### DIFF
--- a/trailblazer/constants.py
+++ b/trailblazer/constants.py
@@ -73,6 +73,10 @@ class SlurmJobStatus(StrEnum):
     def ongoing_statuses(cls) -> tuple:
         return cls.PENDING, cls.RUNNING, cls.COMPLETING
 
+    @classmethod
+    def fail_statuses(cls) -> tuple:
+        return cls.FAILED, cls.TIME_OUT
+
 
 class CharacterFormat(StrEnum):
     """Define character encoding/decoding to use."""
@@ -120,6 +124,10 @@ class TrailblazerStatus(StrEnum):
     @classmethod
     def ongoing_statuses(cls) -> tuple[str, str, str, str]:
         return cls.PENDING, cls.RUNNING, cls.COMPLETING, cls.ERROR
+
+    @classmethod
+    def fail_statuses(cls) -> tuple[str, str, str, str]:
+        return cls.ERROR, cls.FAILED
 
 
 class TrailblazerStatusColor(StrEnum):
--- a/trailblazer/constants.py
+++ b/trailblazer/constants.py
@@ -126,10 +126,6 @@ class TrailblazerStatus(StrEnum):
     def ongoing_statuses(cls) -> tuple[str, str, str, str]:
         return cls.PENDING, cls.RUNNING, cls.COMPLETING, cls.ERROR
 
-    @classmethod
-    def fail_statuses(cls) -> tuple[str, str, str, str]:
-        return cls.ERROR, cls.FAILED
-
 
 class TrailblazerStatusColor(StrEnum):
     """Trailblazer status colors."""
--- a/trailblazer/services/job_service/utils.py
+++ b/trailblazer/services/job_service/utils.py
@@ -1,13 +1,23 @@
 from pathlib import Path
-from trailblazer.constants import FileFormat
+from trailblazer.constants import FileFormat, TrailblazerStatus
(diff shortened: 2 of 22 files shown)


trailblazer/constants.py
```diff
@@ -122,6 +126,10 @@ def statuses(cls) -> tuple:
     def ongoing_statuses(cls) -> tuple[str, str, str, str]:
         return cls.PENDING, cls.RUNNING, cls.COMPLETING, cls.ERROR
 
+    @classmethod
+    def fail_statuses(cls) -> tuple[str, str, str, str]:
+        return cls.ERROR, cls.FAILED
+
```
Vince-janv, line 132: is this not defined on line 77? Also type hint should be `tuple[str, str]`

trailblazer/services/job_service/utils.py
```diff
@@ -1,13 +1,45 @@
...
+def get_status(jobs: list[Job]) -> TrailblazerStatus:
+    is_running: bool = has_running_jobs(jobs)
+    is_failed: bool = has_failed_jobs(jobs)
+    if is_failed and is_running:
+        return TrailblazerStatus.ERROR
+    if is_failed:
+        return TrailblazerStatus.FAILED
+    if is_running:
+        return TrailblazerStatus.RUNNING
+    return TrailblazerStatus.CANCELLED
```
Vince-janv, line 27: Is the old function not removed in this PR? I think there might be one piece of logic missing here:
```
    if single_analysis_status := _get_analysis_single_status(jobs_status_distribution):
        return single_analysis_status
```

trailblazer/services/slurm/slurm_service.py
```diff
@@ -5,5 +5,9 @@
 
 class SlurmService(ABC):
     @abstractmethod
-    def get_job_info(self, job_id: int) -> SlurmJobInfo:
+    def get_job(self, job_id: int) -> SlurmJobInfo:
+        pass
+
+    @abstractmethod
+    def get_jobs(self, job_ids: list[int]) -> list[SlurmJobInfo]:
         pass
```
Vince-janv, line 13: Nice! Looking forward to seeing the REST-API service 😁

trailblazer/store/crud/update.py
```diff
@@ -315,3 +317,16 @@ def update_user_token(self, refresh_token: str, user_id: int) -> None:
         user.refresh_token = refresh_token
         session: Session = get_session()
         session.commit()
+
+    def update_analysis_status(self, analysis_id: int, status: TrailblazerStatus) -> None:
+        analysis: Analysis | None = self.get_analysis_with_id(analysis_id)
+        analysis.status = status
+        session: Session = get_session()
+        session.commit()
```
Vince-janv, line 325: This looks a bit strange, but I guess it's how you do it if you don't pass it as a parameter 🤔

trailblazer/services/job_service/utils.py
seallard: No, I'm leaving removing the deprecated code for a separate PR. It requires untangling a lot of other code.

Added the old logic here https://github.com/Clinical-Genomics/trailblazer/pull/423/commits/cf838778fcad580369e55230caf429304031bdf8.

trailblazer/store/crud/update.py
seallard: I agree, potentially we could just directly import the `Session` and use it as described here https://github.com/Clinical-Genomics/trailblazer/pull/423/commits/cf838778fcad580369e55230caf429304031bdf8.

I'll leave it for now, but might be something we want to revisit.

seallard:
Tested in stage.

seallard:
Deployed to production.

PR #425, opened 2024-04-08
Move rsync filtering

seallard:
## Description
Move filtering of rsync "analyses" into a database filter.

### Fixed
- Move rsync filter into database layer


ChrOertlin (approved):
💯

islean (approved):
Looks good! 💪

PR #426, opened 2024-04-08
(Streamline delivery) Report correct counts

islean:
## Description

Since we will allow users to report even non-completed analyses as delivered, the current way of counting delivered analyses will not work. This PR makes certain that a delivered analysis subtracts from the correct status count.

### Fixed

- Delivered detracts from the correct status count.

### How to prepare for test
- [ ] activate stage: `us`
- [ ] install on stage:
    ```Shell
    ```
- [ ] install on stage:

### How to test
- [ ] login to ...
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] tests executed by

### DIFF
--- a/trailblazer/services/analysis_service/utils.py
+++ b/trailblazer/services/analysis_service/utils.py
@@ -5,31 +5,29 @@ from trailblazer.dto.summaries_response import Summary
 from trailblazer.store.models import Analysis
 
 
-def get_status_count(analyses: list[Analysis], status: TrailblazerStatus) -> int:
-    return len([a for a in analyses if a.status == status])
-
-
-def get_delivered_count(analyses: list[Analysis]) -> int:
-    return len([analysis for analysis in analyses if analysis.delivery])
+def get_status_counts(analyses: list[Analysis]) -> dict[TrailblazerStatus, int]:
+    """Returns the amount of analyses with each status."""
+    delivered: int = 0
+    status_counts: dict = {
+        TrailblazerStatus.CANCELLED: 0,
+        TrailblazerStatus.COMPLETED: 0,
+        TrailblazerStatus.FAILED: 0,
+        TrailblazerStatus.RUNNING: 0,
+    }
+    for analysis in analyses:
+        if analysis.delivery:
+            delivered += 1
+            continue
+        if analysis.status in status_counts.keys():
+            status_counts[analysis.status] += 1
+    status_counts["delivered"] = delivered
+    return status_counts
 
 
 def create_summary(analyses: list[Analysis], order_id: int) -> Summary:
     total: int = len(analyses)
-    completed: int = get_status_count(analyses=analyses, status=TrailblazerStatus.COMPLETED)
-    delivered: int = get_delivered_count(analyses)
-    completed -= delivered
-    running: int = get_status_count(analyses=analyses, status=TrailblazerStatus.RUNNING)
-    cancelled: int = get_status_count(analyses=analyses, status=TrailblazerStatus.CANCELLED)
-    failed = get_status_count(analyses=analyses, status=TrailblazerStatus.FAILED)
-    return Summary(
-        order_id=order_id,
-        total=total,
-        delivered=delivered,
-        running=running,
-        cancelled=cancelled,
-        failed=failed,
-        completed=completed,
-    )
+    status_counts: dict[TrailblazerStatus, int] = get_status_counts(analyses)
+    return Summary(order_id=order_id, total=total, **status_counts)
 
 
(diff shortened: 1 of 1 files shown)


trailblazer/services/analysis_service/utils.py
```diff
@@ -5,31 +5,29 @@
...
+        TrailblazerStatus.CANCELLED: 0,
+        TrailblazerStatus.COMPLETED: 0,
+        TrailblazerStatus.FAILED: 0,
+        TrailblazerStatus.RUNNING: 0,
+    }
+    for analysis in analyses:
+        if analysis.delivery:
+            delivered += 1
+            continue
+        if analysis.status in status_counts.keys():
```
seallard, line 21: Is this check necessary?

trailblazer/services/analysis_service/utils.py
```diff
@@ -5,31 +5,29 @@
...
+            status_counts[analysis.status] += 1
```
islean, line 22: I guess this check is not technically necessary, since we will not render the amount of pending analyses for example. Maybe remove the check? @seallard

seallard: I think it might be necessary due to how the dict is currently constructed. You could consider something like this as well
```python
from collections import defaultdict
from typing import List
from trailblazer.store.models import Analysis, TrailblazerStatus

def get_status_counts(analyses: List[Analysis]) -> dict:
    """Simplifies the counting of analysis statuses including delivered."""
    status_counts = defaultdict(int, {status: 0 for status in TrailblazerStatus})
    status_counts["delivered"] = 0

    for analysis in analyses:
        if analysis.delivery:
            status_counts["delivered"] += 1
        else:
            status_counts[analysis.status] += 1
    if "delivered" in status_counts:
        status_counts[TrailblazerStatus.COMPLETED] -= status_counts["delivered"] 
    return status_counts

def create_summary(analyses: List[Analysis], order_id: int) -> Summary:
    status_counts = get_status_counts(analyses)
    total = len(analyses)
    return Summary(order_id=order_id, total=total, **status_counts)
```

islean: That is probably cleaner, although the check for "delivered" in status_counts will be unnecessary with this way of counting 🥳

seallard (approved):
Nice! You could structure it slightly differently if you want to

trailblazer/services/analysis_service/utils.py
seallard: Yeah, you are right - removed it 👍  Go for which ever implementation you think is better!

islean: Cleaned it up to something more akin to yours! Seems as if you do not need the defaultdict though, list comprehension support for dicts is a thing

trailblazer/services/analysis_service/utils.py
```diff
@@ -5,31 +5,23 @@
...
-def get_status_count(analyses: list[Analysis], status: TrailblazerStatus) -> int:
-    return len([a for a in analyses if a.status == status])
-
-
-def get_delivered_count(analyses: list[Analysis]) -> int:
-    return len([analysis for analysis in analyses if analysis.delivery])
+def get_status_counts(analyses: list[Analysis]) -> dict[TrailblazerStatus, int]:
+    """Returns the amount of analyses with each status."""
+    delivered: int = 0
+    status_counts: dict = {status: 0 for status in TrailblazerStatus}
```
ChrOertlin, line 11: Should delivered be part of TrailblazerStatus? Or are these two separate worlds?

ChrOertlin (approved):
Looks good.

trailblazer/services/analysis_service/utils.py
seallard: Nice, even better!

trailblazer/services/analysis_service/utils.py
islean: That is the million dollar question. Since there was a need for tracking when and by whom it was delivered, it did not fit into the status column. There was also wishes that the TrailblazerStatus only referred to the analysis status, not to the handling after the analysis has been completed. So for now, no, but I am not opposed to a change at some point.

seallard: Sidetrack, `TrailblazerStatus` is a really non descriptive name that creates confusion 😅 It is unclear what it is meant to represent.

islean: Out of scope for this PR though 😅

islean:
Tested on stage by toggling delivered in cigrid-stage. Without it being delivered:

image

When setting to delivered and refreshing:

image

islean:
Deployed to the production VM.

Issue #434, opened 2024-05-06
Bug: tower analysis status update


seallard:
Tower analyses are reported incorrectly as cancelled.

https://github.com/Clinical-Genomics/trailblazer/pull/423 did not reflect that the logic for updating the status of an analysis from its constituent jobs depends on the workflow manager. 

Basically, the Tower statuses differs from the slurm ones and requires a separate translation.

There is also a hard coded assumption that analyses run via Tower has a separate QC step. So once an analysis run in Tower completes, the normal logic does not apply - instead the status of the analysis is set to `QC`. Here is the relevant logic

```python
    @property
    def status(self) -> str:
        """Returns the status of an analysis (also called workflow in NF Tower)."""
        status: str = TOWER_WORKFLOW_STATUS.get(
            self.response.workflow.status, TrailblazerStatus.ERROR
        )

        # If the whole workflow (analysis) is completed set it as QC instead of COMPLETE
        if status == TrailblazerStatus.COMPLETED:
            return TrailblazerStatus.QC
        return status
```

I think this is poor design. It tightly couples the logic of how we define the status of an analysis based on the status of its jobs to the workflow manager used. In the future, are we sure that all analyses via Tower will have a separate QC step like this? Is this how we want to model that?

Alternatives:
Given that the output path is known, can a QC job be scheduled as part of the analysis from the get-go?

seallard:
Quick patch: ensure the status update considers the workflow manager and use `TOWER_WORKFLOW_STATUS.get`.

PR #435, opened 2024-05-06
Ensure old logic is used for updating tower analysis status

seallard:
## Description
Closes https://github.com/Clinical-Genomics/trailblazer/issues/434.
Ensure the old logic is used for updating Tower analysis status and progress.

### Fixed
- Ensure Tower analyses status/progress are reported correctly

### DIFF
--- a/trailblazer/services/analysis_service/analysis_service.py
+++ b/trailblazer/services/analysis_service/analysis_service.py
@@ -1,4 +1,4 @@
-from trailblazer.constants import TrailblazerStatus
+from trailblazer.constants import TrailblazerStatus, WorkflowManager
 from trailblazer.constants import Workflow
 from trailblazer.dto import (
     AnalysesRequest,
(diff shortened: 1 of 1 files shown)


trailblazer/services/analysis_service/analysis_service.py
```diff
@@ -74,6 +74,10 @@ def update_ongoing_analyses(self) -> None:
         analyses: list[Analysis] = self.store.get_ongoing_analyses()
         for analysis in analyses:
             self.job_service.update_jobs(analysis.id)
+
+            if analysis.workflow == WorkflowManager.TOWER:
+                continue
```
seallard, line 79: Could probably move this. Given that the jobs are updated correctly by the job service, I think it might be better to just patch the analysis status update based on the workflow manager

diitaz93 (approved):
Looks good 👍

PR #437, opened 2024-05-06
Fix bug

seallard:
The previous bug fix introduced another bug...

https://github.com/Clinical-Genomics/trailblazer/issues/434


fevac (approved):
🦅 👁️ !

Issue #440, opened 2024-05-13
Status for microsalt analyses reported as cancelled


seallard:
I found that the microsalt cases that started last Friday 2024.5.10 all appear as cancelled in Trailblazer even though they seem to have run fine as there are run directories on hasta and the reports have been emailed. Having status cancelled prevents storing, performing QC and delivering.

seallard:
Logs

```
2024-05-10 22:41:54 hasta.scilifelab.se trailblazer.services.job_service.job_service[59574] ERROR Failed to update jobs case_id - 293385: [Errno 2] No such file or directory: '/home/proj/production/microbial/results/reports/trailblazer/id_slurm_ids.yaml'
2024-05-10 22:41:54 hasta.scilifelab.se trailblazer.store.crud.update[59574] INFO Updated status cas_id - 293385: cancelled 
```

PR #442, opened 2024-05-13
Fix job status update

seallard:
## Description
Closes https://github.com/Clinical-Genomics/trailblazer/issues/440
Ensure exception is raised when job service fails updating jobs. The caller (analysis service) will catch the exception and handle it correctly.

### Fixed
- Ensure job service raises exception when it fails

### DIFF
--- a/trailblazer/services/job_service/job_service.py
+++ b/trailblazer/services/job_service/job_service.py
@@ -3,7 +3,7 @@ import logging
 
 from trailblazer.constants import TrailblazerStatus, WorkflowManager
 from trailblazer.dto import CreateJobRequest, FailedJobsRequest, FailedJobsResponse, JobResponse
-from trailblazer.exceptions import JobServiceError
+from trailblazer.exceptions import JobServiceError, NoJobsError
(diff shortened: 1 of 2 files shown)


trailblazer/services/job_service/job_service.py
```diff
@@ -48,6 +49,7 @@ def update_jobs(self, analysis_id: int) -> None:
                 self.store.update_tower_run_status(analysis_id)
         except Exception as error:
             LOG.error(f"Failed to update jobs {analysis.case_id} - {analysis.id}: {error}")
+            raise JobServiceError from error
```
islean, line 52: How does this help in preventing the problematic behaviour? Will they show up as "failed" instead of "cancelled"?

seallard: The old logic will ensure the analysis status is set to `ERROR` when an exception bubbles up. The problem was that the exception was not propagated from the job service.

islean (approved):
Cryptic microsalt but I guess this should fix it

PR #444, opened 2024-05-14
Remove action duplication

seallard:
## Description
The action building and publishing images on pushes are somewhat duplicated for the web API and CLI images. This PR consolidates them.


ChrOertlin (approved):
If the order in which they were run does not matter, looks good to me

seallard:
> If the order in which they were run does not matter, looks good to me

It does not matter 👍

PR #445, opened 2024-05-15
Remove crossplatform option

seallard:
## Description
The containerization software we run on the cluster (Singularity) cannot run images generated with the `buildx` option. It throws errors like `unsupported schema version 2`. This is due to the software being outdated by 5 years.

The issue is resolved by removing buildx (it lets you build images for both arm64 and amd64 architectures). We only use `amd64` on our servers.

The only drawback is that our Macs use arm64, meaning that the images will not be possible to run on our local machines. I can sidestep this by reintroducing the duplication and have a separate one for the CLI images intended to run on the cluster. I'm fine with either option. I don't think anyone in our team uses the web API image locally anyway.

### Fixed
- Remove buildx option breaking old containerization software on cluster

### DIFF
--- a/.github/workflows/build_dockerfile_on_push.yml
+++ b/.github/workflows/build_dockerfile_on_push.yml
@@ -20,10 +20,6 @@ jobs:
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}
 
-     - name: Set up Docker Buildx
-       id: buildx
-       uses: docker/setup-buildx-action@v2.2.1
-
      - name: Build web API image
        uses: docker/build-push-action@v3.2.0
        with:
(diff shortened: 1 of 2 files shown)


Vince-janv (approved):
I don't think the container needs to be able to run on our macs. I'm a bit confused though. I can't find any arm64 containers on dockerhub. Do we not push them?

seallard:
> I don't think the container needs to be able to run on our macs. I'm a bit confused though. I can't find any arm64 containers on dockerhub. Do we not push them?

I dug around and you are correct, is seems like the default `--platform` option with buildx is taken from the environment the builds is done in. So currently, we only build for `linux/amd64` if I understand the [docs](https://docs.docker.com/reference/cli/docker/buildx/build/#platform) correctly.

PR #447, opened 2024-05-16
Update deployment instructions

seallard:
Update deployment instructions to reflect how to use the CLI images on Hasta.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -23,7 +23,7 @@ Alembic revisions need to be applied from your local machine.
 
 
 # Steps when merging to master
-When all tests done and successful and PR is approved, follow these steps.
+When all the tests are done and successful and the PR is approved, follow these steps.
 
 1. Select "Squash and merge" to merge branch into default branch (master/main).
 2. A prompt for writing merge commit message will pop up.
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -1,6 +1,6 @@
 # Deployment for testing
-Comment on the PR with the deploy output, including the environment and that it succeeded.
-Alembic revisions need to be applied from your local machine.
+Comment on the PR with the deploy output, it should include environment and that it succeeded.
+Alembic revisions need to be applied from your local machine. See Atlas for instructions.
 
 ## Deploy CLI feature branch
 1. `ssh hasta`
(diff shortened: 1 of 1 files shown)


DEPLOYMENT.md
```diff
@@ -1,49 +1,35 @@
-# Steps
+# Deployment for testing
+Comment on the PR with the deploy output, it should include environment and that it succeeded.
+Alembic revisions need to be applied from your local machine.
```
ChrOertlin, line 3: Instructions on how this is done would be good. If someone has not worked with alembic before this can be quite cumbersome to figure out.

DEPLOYMENT.md
```diff
@@ -1,49 +1,35 @@
-# Steps
+# Deployment for testing
+Comment on the PR with the deploy output, it should include environment and that it succeeded.
```
diitaz93, line 2:
```suggestion
Comment on the PR with the deploy output, including the environment and that it succeeded.
```

ChrOertlin (approved):
Looks good, instructions on how to perform alembic migrations from a local machine would be great

DEPLOYMENT.md
```diff
@@ -1,49 +1,35 @@
...
+3. `trailblazer-deploy <version>`
 
+## Deploy web API
+1. Use the `Deploy release to production environment` Github action.
 
-3. Find the title of the pull request already pre-filled in the merge commit title, or copy and paste 
-the title if not.
 
+# Steps when merging to master
+When all tests done and successful and PR is approved, follow these steps.
```
diitaz93, line 26:
```suggestion
When all the tests are done and successful and the PR is approved, follow these steps.
```

diitaz93 (approved):
👍

Vince-janv (approved):
Nice 👍

DEPLOYMENT.md
seallard: Referenced Atlas.

seallard:
Had to give up on this. Basically, when running the image you need to mount slurm from the host (to be able to use `squeue`). However, to authenticate you need access to the `munge` config on the host as well, specifically the socket and key. You need sudo for that, they cannot be mounted. Which means we cannot retrieve info about the slurm jobs from inside the container.

If we had somewhat more recent version of SLURM, we could retrieve the info via the slurm rest API instead.

PR #450, opened 2024-06-05
Update base image

seallard:
Switch to UBI base image.


seallard:
Not sure if this is what we want

PR #451, opened 2024-06-05
Add slurm api service

seallard:
## Description
Add and conditionally instantiate slurm API service to retrieve information about slurm jobs.

Tested on the hospitals cluster.


trailblazer/containers.py
```diff
@@ -20,6 +22,9 @@ class Container(containers.DeclarativeContainer):
     google_oauth_base_url: str = os.environ.get("GOOGLE_OAUTH_BASE_URL")
     encryption_key: str = os.environ.get("ENCRYPTION_KEY")
     google_api_base_url: str = os.environ.get("GOOGLE_API_BASE_URL")
+    slurm_jwt_token: str = os.environ.get("SLURM_JWT")
+    slurm_user_name: str = os.environ.get("SLURM_USER_NAME")
+    slurm_base_url: str = os.environ.get("SLURM_BASE_URL")
```
Vince-janv, line 27: What happens if these do not exist? Are the variables just empty strings or do we need to add these in our current configs?

trailblazer/containers.py
```diff
@@ -32,8 +37,17 @@ class Container(containers.DeclarativeContainer):
...
-    slurm_client = providers.Singleton(SlurmCLIClient, host=slurm_host)
-    slurm_service = providers.Singleton(SlurmCLIService, client=slurm_client)
+
+    if slurm_base_url and slurm_jwt_token and slurm_user_name:
+        slurm_client = providers.Singleton(
+            SlurmAPIClient,
+            base_url=slurm_base_url,
+            access_token=slurm_jwt_token,
+            user_name=slurm_user_name,
+        )
```
Vince-janv, line 47: I guess this answers my previous question 😅

trailblazer/services/slurm/slurm_api_service/slurm_api_service.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SlurmAPIService(SlurmService):
+    def __init__(self, client: SlurmAPIClient):
+        self.client = client
+
+    def get_job(self, job_id: int) -> SlurmJobInfo | None:
+        job_response: SlurmJobResponse = self.client.get_job(job_id)
+        if not job_response.jobs or job_response.errors:
+            return None
```
Vince-janv, line 15: Are we leaving error handling for the future?

trailblazer/services/slurm/slurm_api_service/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        elapsed=elapsed,
+    )
+
+
+def get_job_elapsed_time(job: SlurmAPIJobInfo) -> int:
+    return job.accrue_time.number if job.accrue_time and job.accrue_time.set else 0
+
+
+def get_job_start_time(job: SlurmAPIJobInfo) -> datetime | None:
+    if job.start_time is None or job.start_time.set is None:
```
Vince-janv, line 28: What do the `.set` fields do?

trailblazer/services/slurm/slurm_api_service/slurm_api_service.py
```diff
@@ -0,0 +1,24 @@
...
+        return create_job_info_dto(job_response)
+
+    def get_jobs(self, job_ids: list[int]) -> list[SlurmJobInfo]:
+        jobs: list[SlurmJobInfo] = []
+        for job_id in job_ids:
+            job: SlurmJobInfo | None = self.get_job(job_id)
+            if job:
+                jobs.append(job)
```
Vince-janv, line 23: If we want to make this efficient I think we might want to change the logic. I think it might be faster to iterate over all list and check if they exist in job_ids rather than iterating through the small one and referencing the big one.

Feel free to leave this since we're not using this function atm

Vince-janv (approved):
Love the dependency injection! But is there a better way than checking env vars in if-statements to decide what to instantiate?

trailblazer/containers.py
seallard: Ah, good point. The type hint is wrong, it is a string or none. So no, we do not need to update our configs

seallard: Fixed type hints in https://github.com/Clinical-Genomics/trailblazer/pull/451/commits/a4d97470c85cca02a7e045d2ab7e38a2670b52b8

trailblazer/services/slurm/slurm_api_service/utils.py
seallard: That is a field provided by the slurm API in the response to indicate whether the field has been set. I think it is redundant and I'll remove it

trailblazer/services/slurm/slurm_api_service/slurm_api_service.py
seallard: I don't understand this suggestion! Do you mean that we should retrieve information about all jobs currently pending/running by SLURM and use that list?

That could potentially be more efficient since it would decrease the number of requests. I don't think this part of the code is performance critical though, so optimizing it now seems unnecessary - especially since it is not used like you say.

Vince-janv: Yeah, I meant fetching all jobs and then iterating through them and picking out the jobs in job_ids.

Let's skip it for now

seallard: I think that implementation makes more sense, but lets leave it for now

trailblazer/services/slurm/slurm_api_service/utils.py
seallard: Removed in https://github.com/Clinical-Genomics/trailblazer/pull/451/commits/81b0af1a637b1123def6fa402758fbdc4c52e72e

trailblazer/services/slurm/slurm_api_service/slurm_api_service.py
seallard: Good point, to ensure the behavior that we have for the `SlurmCLIService` we should raise an exception the client instead for this. The existing behaviour is that, if a single job cannot be updated for an analysis, the entire status of the analysis is set to error.

seallard:
> Love the dependency injection! But is there a better way than checking env vars in if-statements to decide what to instantiate?

I have not figured out a better way 🤔

trailblazer/services/slurm/slurm_api_service/slurm_api_service.py
seallard: This already happens. Removed the unnecessary check and added some additional error handling in https://github.com/Clinical-Genomics/trailblazer/pull/451/commits/3d452a77bff25502e071e61a96d1af9112906dbb

Issue #436, opened 2024-05-06
Decouple Tower logic from database layer


seallard:
`update.py` contains logic related to the Tower API. It should be extracted from there.
The Tower API is also instantiated per analysis. It should be refactored to only be instantiated once, it should also be split into two:
- A TowerAPIClient which only handles the interactions with the REST API
- A TowerService which wraps around the client, doing any necessary additional things

PR #463, opened 2024-07-23
Clean up tower and slurm interactions

seallard:
Closes https://github.com/Clinical-Genomics/trailblazer/issues/436

This PR does the following
- Standardize the Tower API client
- Standardize the Tower API service
- Extract all tower and slurm interactions from the data access layer in `update.py` to the job service.
- Rewrite tests depending on the old service, client and `update.py` module.

Sadly, it was difficult to split this into smaller chunks because everything was tangled up. Most of the code change consists of tests though 😄 

The tower client and corresponding service is designed in an uncommon way violating some design principles.
This PR refactors the client as to follow the design we have established in [Atlas](https://atlas.scilifelab.se/infrastructure/development/cgep/cgep_6_consume_web_apis/). It also refactors the corresponding Tower service to make use of the client as expected and extracts the Tower interactions from the data access layer.


trailblazer/services/job_service/job_service.py
```diff
@@ -58,13 +67,26 @@ def _update_slurm_jobs(self, analysis_id: int) -> None:
...
 
         if not analysis.jobs:
             raise NoJobsError(f"No jobs found for analysis {analysis_id}")
 
-        return get_status(analysis.jobs)
+        status: TrailblazerStatus = get_status(analysis.jobs)
+        if (
+            analysis.workflow_manager == WorkflowManager.TOWER
+            and status == TrailblazerStatus.COMPLETED
+        ):
```
seallard, line 87: Extract conditional

trailblazer/services/job_service/job_service.py
```diff
@@ -58,13 +67,26 @@ def _update_slurm_jobs(self, analysis_id: int) -> None:
         jobs: list[Job] = [slurm_info_to_job(job_info) for job_info in slurm_jobs]
         self.store.replace_jobs(analysis_id=analysis_id, jobs=jobs)
 
+    def _update_tower_jobs(self, analysis_id: int) -> None:
+        analysis: Analysis = self.store.get_analysis_with_id(analysis_id)
+        workflow_id: str = get_tower_workflow_id(analysis)
+        tower_jobs: list[TowerJobInfo] = self.tower_service.get_jobs(workflow_id)
```
seallard, line 73: This is kind of janky. Would be nice to encapsulate the logic for retrieving the workflow id in the tower service. Same goes for the slurm job ids.

trailblazer/clients/slurm_api_client/slurm_api_client.py
```diff
@@ -31,3 +28,12 @@ def get_job(self, job_id: str) -> SlurmJobResponse:
...
             raise ResponseDeserializationError(e)
+
+    def cancel_job(self, job_id: str) -> None:
+        endpoint: str = f"{self.base_url}/slurm/v0.0.40/job/{job_id}"
+        try:
+            response = requests.delete(endpoint, headers=self.headers)
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            LOG.error(f"Error cancelling job {job_id}: {e.response.content}")
+            raise SlurmAPIClientError(e)
```
Vince-janv, line 39: Was this functionality missing from the slurm_api_client before?

trailblazer/clients/tower/models.py
```diff
@@ -94,6 +94,10 @@ def is_complete(cls) -> bool:
         return cls.status == SlurmJobStatus.COMPLETED
 
 
+class TaskWrapper(BaseModel):
+    task: TowerTask
```
Vince-janv, line 98: Why is a wrapper needed?

trailblazer/clients/slurm_api_client/slurm_api_client.py
```diff
@@ -31,3 +28,12 @@ def get_job(self, job_id: str) -> SlurmJobResponse:
         except ValidationError as e:
             LOG.error(f"Error deserializing job response: {e}")
             raise ResponseDeserializationError(e)
+
+    def cancel_job(self, job_id: str) -> None:
+        endpoint: str = f"{self.base_url}/slurm/v0.0.40/job/{job_id}"
```
fevac, line 33: I don't understand why this version is fixed in the code (v0.0.40) and not retrieved from somewhere else or stored as a constant if needed.

trailblazer/clients/tower/tower_client.py
```diff
@@ -1,128 +1,43 @@
...
+        response = requests.get(url=url, headers=self.headers, params=self.request_params)
+        response.raise_for_status()
+        json = response.json()
+        return TowerWorkflowResponse.model_validate(json)
+
+    @handle_client_errors
+    def cancel_workflow(self, workflow_id: str) -> None:
+        url = f"{self.base_url}/workflow/{workflow_id}/cancel"
+        response = requests.post(url=url, headers=self.headers, params=self.request_params, json={})
+        response.raise_for_status()
```
Vince-janv, line 43: great clean up 👍

trailblazer/clients/tower/utils.py
```diff
@@ -0,0 +1,27 @@
...
+            LOG.error(f"Request against tower failed: {error}")
+            raise TowerRequestFailed(error) from error
+        except ValidationError as error:
+            LOG.error(f"Invalidly formatted response from Tower API: {error}")
+            raise InvalidTowerAPIResponse(error) from error
+        except Exception as error:
+            LOG.error(f"Unexpected error in Tower API client: {error}")
+            raise TowerAPIClientError(error) from error
+
+    return wrapper
```
Vince-janv, line 27: Fancy 🧐

trailblazer/clients/tower/tower_client.py
```diff
@@ -1,128 +1,43 @@
...
-    def headers(self) -> dict:
-        """Return headers required for an NF Tower API call.
-        Accept and Authorization fields are mandatory."""
-        return {
+    def __init__(self, base_url: str, access_token: str, workspace_id: str):
+        self.base_url = base_url
+        self.access_token = access_token
+        self.workspace_id = workspace_id
+        self.request_params = [("workspaceId", self.workspace_id)]
+        self.headers = {
```
fevac, line 18: 🌟

trailblazer/services/job_service/job_service.py
```diff
@@ -64,8 +71,24 @@ def get_analysis_status(self, analysis_id: int) -> TrailblazerStatus:
...
             raise NoJobsError(f"No jobs found for analysis {analysis_id}")
 
-        return get_status(analysis.jobs)
+        status: TrailblazerStatus = get_status(analysis.jobs)
+        if (
+            analysis.workflow_manager == WorkflowManager.TOWER
+            and status == TrailblazerStatus.COMPLETED
+        ):
+            status = TrailblazerStatus.QC
+        return status
```
Vince-janv, line 80: Where is this logic coming from?

trailblazer/clients/tower/utils.py
```diff
@@ -0,0 +1,27 @@
...
+from pydantic import ValidationError
+from requests import ConnectionError, HTTPError
+from requests.exceptions import MissingSchema
+
+from trailblazer.exc import InvalidTowerAPIResponse, TowerAPIClientError, TowerRequestFailed
+
+LOG = logging.getLogger(__name__)
+
+
+def handle_client_errors(func):
```
fevac, line 12: Add docstring and type hint

trailblazer/clients/tower/tower_client.py
```diff
@@ -1,128 +1,43 @@
...
-        if self.meets_requirements:
-            url = self.build_url(endpoint=self.workflow_endpoint)
-            return TowerWorkflowResponse(**self.send_request(url=url))
-
-    def send_cancel_request(self) -> None:
-        """Send a POST request to cancel a workflow."""
-        if self.meets_requirements:
-            url: str = self.build_url(endpoint=self.cancel_endpoint)
-            self.post_request(url=url)
+    @handle_client_errors
```
fevac, line 23: ohh nice! love this decorator

trailblazer/services/job_service/job_service.py
```diff
@@ -46,7 +53,7 @@ def update_jobs(self, analysis_id: int) -> None:
             if analysis.workflow_manager == WorkflowManager.SLURM:
                 self._update_slurm_jobs(analysis_id)
             if analysis.workflow_manager == WorkflowManager.TOWER:
-                self.store.update_tower_run_status(analysis_id)
+                self.tower_service.update_jobs(analysis_id)
```
fevac, line 56: just my ignorance here, but why the syntax is different for slurm and tower? meaning, why _update_slurm_jobs (starting with _) but update_jobs for tower (without)?

trailblazer/services/job_service/job_service.py
fevac, line 80: For the future, we could use `workflow.onComplete` and `workflow.onError` statements within tower itself. This allow to trigger something or run some code after the pipeline is succesful or if it exists with errors. We could transfer the logic there if that's preferred and set the trailblazer status on pipeline completion. Anyways, this is not part of this PR and I'm not sure we want that either (would need to be thought through)

tests/store/crud/test_update.py
Vince-janv, line 1: If I'm not mistaken, we're removing more tests than we're adding. Did you deem some tests unnecessary?

Vince-janv (approved):
Nice! Well done ⭐ Have these changes been tested in stage?

trailblazer/services/tower/utils.py
```diff
@@ -0,0 +1,21 @@
...
+from trailblazer.store.models import Analysis, Job
+
+
+def get_tower_workflow_id(analysis: Analysis) -> str:
+    file = Path(analysis.config_path)
+    content: dict = ReadFile.get_content_from_file(file_format=FileFormat.YAML, file_path=file)
+    return content.get(analysis.case_id)[-1]
+
+
+def create_job_from_dto(task: TowerTask) -> Job:
```
fevac, line 14: docstring? I don't know what dto stands for 😅

fevac (commented):
Nice clean up!! 🌟  Leaving some comments but I'm not the most technical person, so it would be good with extra eyes

tests/store/crud/test_update.py
seallard: True! Some of the functions being tested no longer exists, which is why they have been removed.
I'm writing tests currently for the refactored logic. They will cover the analysis and job service.

I'll ping you when it is done and this comment can be resolved.

trailblazer/clients/slurm_api_client/slurm_api_client.py
seallard: Good point, it should be in the base url 👍

trailblazer/services/job_service/job_service.py
seallard: Yes, good point! I'll make sure to refactor it to look the same for slurm but I don't want to do more stuff in this PR 😅

trailblazer/services/job_service/job_service.py
seallard: I dug it out from the old code: https://github.com/Clinical-Genomics/trailblazer/blob/daa140fca1743db625a7092fd772453c7e91031f/trailblazer/services/tower/tower_api_service.py#L62

All workflows that run in tower have a separate QC step which is why when the analysis completes the status is not set to `completed` but rather `qc`. I don't think it is a good design and it does not seem reasonable to assume that all workflows using tower will have a separate QC step.

I don't want to fix that in this PR.

trailblazer/clients/tower/models.py
seallard: The response data from tower is structured that way: https://cloud.seqera.io/openapi/#get-/workflow/-workflowId-/tasks.

We could hide it away by doing some pydantic magic, but I rather keep the structure close to the actual response structure to limit confusion. I did add a `get_tasks` utils method to the `TowerTasksResponse` model to discard the wrapper and just get the tasks.

trailblazer/clients/slurm_api_client/slurm_api_client.py
seallard: Yes!

trailblazer/services/job_service/job_service.py
seallard: That is nice! I think that could make sense, unless we want to assume that all workflows run via Tower in the future always will have a separate QC step.

Alternatively, we could track the QC job directly on the analysis in trailblazer somehow.

trailblazer/clients/tower/models.py
Vince-janv: oof, if the other option is `tasks: list[dict[str, TowerTask]]` then i prefer this 😅

trailblazer/services/job_service/job_service.py
Vince-janv: Agreed, that we shouldn't touch the logic. I just wanted to make sure that it wasn't something you cooked up yourself 😉

trailblazer/clients/slurm_api_client/slurm_api_client.py
seallard: I completely agree with this comment, but I'll save it for another PR to limit the headaches in this one

trailblazer/clients/tower/utils.py
seallard: Added docstring here https://github.com/Clinical-Genomics/trailblazer/pull/463/commits/e688385e7f245d3a6786b0e9fa5a5d36c91f4f87

The type hint for the function would not add any clarity and I will omit it

trailblazer/services/tower/utils.py
seallard: Renamed function in https://github.com/Clinical-Genomics/trailblazer/pull/463/commits/153c631d09c1ddc2285258574837beced185ed71

tests/store/crud/test_update.py
seallard: Tests for new tower client:
Tests for new tower service:
Tests for job service:
Tests for analysis service:
Integration tests for analysis service:
Integration tests for CLI commands:

seallard:
To test in stage
- [ ] scan updates tower analyses correctly
- [ ] scan updates slurm analyses correctly
- [ ] cancel tower analysis
- [ ] cancel slurm analysis

tests/store/crud/test_update.py
seallard: I found a bug in the slurm CLI service, I should add tests for that service alternatively include it in an integration test.

trailblazer/store/crud/update.py
```diff
@@ -41,167 +30,11 @@ def update_user_is_archived(self, user: User, archive: bool = True) -> None:
...
-            LOG.error(f"Error logging case - {analysis.case_id} :  {type(error).__name__}")
-            analysis.status = TrailblazerStatus.ERROR
-            session: Session = get_session()
-            session.commit()
-
-    def _update_analysis_from_tower_output(
-        self, analysis: Analysis, analysis_id: int, tower_api: TowerAPIService
-    ):
-        LOG.info(f"Status in Tower: {analysis.case_id} - {analysis_id} - {tower_api.workflow_id}")
-        analysis.status = tower_api.status
```
seallard, line 137: The refactored version calculates this status from the jobs persisted in the database for the analysis.

It might be the case that tower differs from slurm in how jobs are created and therefor this might be wrong.
If new jobs can be added when all the previous ones completed, the status will be incorrectly reported as completed.

trailblazer/services/tower/tower_api_service.py
```diff
@@ -1,148 +1,26 @@
...
-        pending, running, completed or failed."""
-        if not self._tasks_response:
-            self._tasks_response = self.tower_client.tasks
-        return self._tasks_response
-
-    @property
-    def status(self) -> str:
-        """Returns the status of an analysis (also called workflow in NF Tower)."""
-        status: str = TOWER_WORKFLOW_STATUS.get(
-            self.response.workflow.status, TrailblazerStatus.ERROR
```
seallard, line 59: The overall status for tower analyses seems to be mirrored by the workflow status, no its constituent jobs.

trailblazer/services/tower/tower_api_service.py
```diff
@@ -1,148 +1,26 @@
...
-    def response(self) -> TowerWorkflowResponse:
-        """Returns a workflow response containing general information about an analysis."""
-        if not self._response:
-            self._response = self.tower_client.workflow
-        return self._response
-
-    @property
-    def tasks_response(self) -> TowerTaskResponse:
-        """Returns a tasks response containing information about jobs submitted as part of an analysis.
-        A task response does not include all jobs that will be run as part of an analysis.
```
seallard, line 48: This is the reason the analysis status calculation is incorrect

PR #464, opened 2024-07-24
Require user abbreviation

seallard:
Mark user abbreviation as non nullable. Part of resolving https://github.com/Clinical-Genomics/cigrid-ui/issues/600.


seallard:
Applied in stage.

seallard:
Applied and deployed to prod.

PR #467, opened 2024-07-26
Set upload date for FASTQ analysis

seallard:
Closes https://github.com/Clinical-Genomics/cigrid-ui/issues/655.

### DIFF
--- a/trailblazer/services/analysis_service/utils.py
+++ b/trailblazer/services/analysis_service/utils.py
@@ -1,8 +1,10 @@
-from trailblazer.constants import TrailblazerStatus
+from datetime import datetime, timedelta
+from trailblazer.constants import SlurmJobStatus, TrailblazerStatus
 from trailblazer.dto.analyses_response import UpdateAnalysesResponse
 from trailblazer.dto.analysis_response import AnalysisResponse
 from trailblazer.dto.summaries_response import StatusSummary, Summary
 from trailblazer.store.models import Analysis
+from trailblazer.store.models import Job as DbJob
 
 
 def get_status_counts(analyses: list[Analysis]) -> dict[TrailblazerStatus, StatusSummary]:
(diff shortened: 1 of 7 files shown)


ChrOertlin (approved):
Perhaps a more informative name for DbJob, looks good otherwise!

trailblazer/services/analysis_service/utils.py
```diff
@@ -1,8 +1,10 @@
-from trailblazer.constants import TrailblazerStatus
+from datetime import datetime, timedelta
+from trailblazer.constants import SlurmJobStatus, TrailblazerStatus, Workflow
 from trailblazer.dto.analyses_response import UpdateAnalysesResponse
 from trailblazer.dto.analysis_response import AnalysisResponse
 from trailblazer.dto.summaries_response import StatusSummary, Summary
 from trailblazer.store.models import Analysis
+from trailblazer.store.models import Job as DbJob
```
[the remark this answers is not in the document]
seallard: True, also unnecessary! https://github.com/Clinical-Genomics/trailblazer/pull/467/commits/d4533ec7d779d7a9c4050c57059102670c533185

trailblazer/services/analysis_service/utils.py
```diff
@@ -41,3 +43,24 @@ def create_update_analyses_response(analyses: list[Analysis]) -> UpdateAnalysesR
...
+
+
+def _get_completed_job(jobs: list[Job]) -> Job | None:
+    for job in jobs:
+        if _is_completed_job(job):
+            return job
+
+
+def _is_completed_job(job: Job):
+    return job.status == SlurmJobStatus.COMPLETED and job.started_at and job.elapsed
```
seallard, line 62: Sometimes the elapsed time is 0, make this an explicit null check

trailblazer/services/analysis_service/utils.py
```diff
@@ -41,3 +43,24 @@ def create_update_analyses_response(analyses: list[Analysis]) -> UpdateAnalysesR
...
+
+
+def _get_completed_at_date(job: Job) -> datetime:
+    return job.started_at + timedelta(seconds=job.elapsed)
```
seallard, line 66: Most upload jobs finish within minutes, so you could argue that this is irrelevant given that we just want the day that the job

seallard (commented):
The approach needs to be changed, the way the logic is hooked up will not work.

The logic is run for ongoing analyses. The upload only happens after the analysis is completed, so the logic needs to be shifted elsewhere.

seallard:
Tested in stage

PR #469, opened 2024-07-29
Filter on workflow for upload status logic

seallard:
Small cleanup, more efficient to filter out any non-fastq analyses earlier on since those are the only analyses we want to set the upload date for automatically currently.

### DIFF
--- a/trailblazer/services/analysis_service/utils.py
+++ b/trailblazer/services/analysis_service/utils.py
@@ -47,7 +47,7 @@ def create_update_analyses_response(analyses: list[Analysis]) -> UpdateAnalysesR
 
 def get_upload_date(analysis: Analysis) -> datetime | None:
     completed_job: Job | None = _get_completed_job(analysis.upload_jobs)
-    if analysis.workflow != Workflow.FASTQ or not completed_job:
+    if not completed_job:
         return
     return _get_completed_at_date(completed_job)
 
--- a/trailblazer/store/crud/read.py
+++ b/trailblazer/store/crud/read.py
@@ -4,7 +4,7 @@ from typing import Callable
 from sqlalchemy import desc
 from sqlalchemy.orm import Query
 
-from trailblazer.constants import JobType, TrailblazerStatus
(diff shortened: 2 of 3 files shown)


trailblazer/services/analysis_service/utils.py
```diff
@@ -47,7 +47,7 @@ def create_update_analyses_response(analyses: list[Analysis]) -> UpdateAnalysesR
 
 def get_upload_date(analysis: Analysis) -> datetime | None:
     completed_job: Job | None = _get_completed_job(analysis.upload_jobs)
```
islean, line 49: Not for this PR to solve, but I it is slightly annoying that analysis.upload_jobs is typehinted as singular here. Is it plural or singular?

islean (approved):
LGTM!

trailblazer/store/crud/read.py
```diff
@@ -253,8 +253,13 @@ def _paginate_analyses(self, analyses: Query, request: AnalysesRequest) -> Query
...
         return apply_analysis_filter(
-            filter_functions=[AnalysisFilter.BY_NOT_UPLOADED, AnalysisFilter.BY_COMPLETED],
+            filter_functions=[
+                AnalysisFilter.BY_NOT_UPLOADED,
+                AnalysisFilter.BY_COMPLETED,
+                AnalysisFilter.BY_WORKFLOW,
+            ],
             analyses=self.get_query(Analysis),
+            workflow=Workflow.FASTQ,
         ).all()
```
ChrOertlin, line 265: can we make this a general function that receives a workflow?

seallard: Sure 👍 https://github.com/Clinical-Genomics/trailblazer/pull/469/commits/8a0a8b125d3e38518604b8bc6879b8c6f453951b

trailblazer/services/analysis_service/utils.py
seallard: It is plural, but the function just retrieves one of the completed upload jobs.

Usually there is only one upload job, but there can also be several failed ones for FASTQ

PR #471, opened 2024-08-07
Remove job step formatting logic

seallard:
We have a lot of logic for reformatting something called a "job step" from slurm.

This value is not stored anywhere or used anywhere in the logic for anything. So we can remove it 😅


seallard:
Nvm, the step is actually used as the name of the job 😢

PR #474, opened 2024-08-16
Add endpoint error handling

seallard:
This PR removes a lot of duplicated error handling logic in the endpoints by extracting it to a separate error handler. 

The code is easier to read since the indentation is decreased, duplicated logic is removed, fewer overall lines of code and improved separation of concerns 👍 

Additionally, it:
- Ensures we do not return stack traces in the error response (security risk) but just log them instead.
- Ensures the correct HTTP status is returned for internal server errors (should not be 409 for a blank catch-all exception).


trailblazer/server/utils.py
```diff
@@ -25,3 +33,26 @@ def stringify_timestamps(data: dict) -> dict[str, str]:
...
+        except MissingAnalysis as error:
+            return jsonify(error=str(error)), HTTPStatus.NOT_FOUND
+        except ValidationError as error:
+            LOG.error(f"Validation error in analysis endpoint: {error}")
+            return jsonify(error=str(error)), HTTPStatus.BAD_REQUEST
+        except Exception as error:
+            LOG.error(f"Unexpected error in analysis endpoint: {error}")
+            return (
+                jsonify(error="An error occurred while processing your request."),
+                HTTPStatus.INTERNAL_SERVER_ERROR,
```
islean, line 55: Do we have any logic depending on the Status being CONFLICT here?

ChrOertlin (approved):
Nice job!

trailblazer/server/utils.py
seallard: I went through the cg client and the cigrid web app and could not find any such logic depending on the status code being 409

PR #478, opened 2024-08-23
Patch tower analysis status update

seallard:
We prematurely set the status of tower analyses to QC due to not taking into account that all jobs in a workflow can be completed without the workflow itself being completed.


Vince-janv (approved):
Nice, this logic looks good 👍

PR #479, opened 2024-08-26
Add analysis in cancel response

seallard:
Include the updated analysis in the response from the `/cancel` endpoint.


Vince-janv (approved):
👍

PR #480, opened 2024-08-26
Extract slurm client error handling

seallard:
The error handling does not belong in the client. This PR extracts it to a separate wrapper.

### DIFF
--- a/trailblazer/clients/slurm_api_client/error_handler.py
+++ b/trailblazer/clients/slurm_api_client/error_handler.py
@@ -2,7 +2,8 @@ from functools import wraps
 import logging
 
 from pydantic import ValidationError
-import requests
+from requests import HTTPError
+
 
 from trailblazer.exc import ResponseDeserializationError, SlurmAPIClientError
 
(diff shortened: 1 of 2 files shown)


trailblazer/clients/slurm_api_client/error_handler.py
```diff
@@ -0,0 +1,25 @@
...
+LOG = logging.getLogger(__name__)
+
+
+def handle_errors(func):
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except requests.exceptions.HTTPError as e:
```
islean, line 18:
```suggestion
        HTTPError as e:
```
According to convention we do not really import whole packages

trailblazer/clients/slurm_api_client/error_handler.py
```diff
@@ -0,0 +1,25 @@
...
+            LOG.error(f"Error getting job: {e.response.content}")
```
islean, line 19: This logging is a bit misguiding since it would be the logging shown when a "cancel job" request fails. Previous logging in that instance was:
"LOG.error(f"Error cancelling job {job_id}: {e.response.content}")"

islean (approved):
Some minor comments but looks good as a whole!

PR #481, opened 2024-08-26
Add tower service error handler

seallard:


trailblazer/services/tower/error_handler.py
```diff
@@ -0,0 +1,21 @@
...
+
+def handle_errors(func):
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except FileNotFoundError as error:
+            LOG.error(f"File not found: {error}")
+            raise WorkflowIdFileMissingError()
```
islean, line 17: Seems a bit brittle for the future to me, say if it could be another file missing. Would it not make sense that a more specific error is returned already when the workflow id file is missing?

islean (approved):
Looks good although the error handling might be a bit brittle? A bit unsure of what is best.

trailblazer/services/tower/error_handler.py
seallard: Yes, if it in the future becomes possible for another file to be missing in the Tower API service it would have to be changed. Currently, that is not the case and as such I don't think it is a concern!

PR #483, opened 2024-08-26
Add tower workflow id to analysis

seallard:
Part of resolving https://github.com/Clinical-Genomics/cigrid-ui/issues/672

This PR adds a tower workflow id column to the analysis table in trailblazer. It is really inconvenient and messy to try to read the file with the id from Hasta in the VM.

After the refactoring https://github.com/Clinical-Genomics/trailblazer/pull/463, the tower and slurm logic has been separated cleanly into different modules and there is no need to resort to the slurm shenanigans for tower analyses.

### DIFF
--- a/trailblazer/store/models.py
+++ b/trailblazer/store/models.py
@@ -92,6 +92,7 @@ class Analysis(Model):
     version = Column(types.String(32))
     workflow = Column(types.String(32))
     workflow_manager = Column(types.Enum(*WorkflowManager.list()), default=WorkflowManager.SLURM)
+    tower_workflow_id = Column(types.String(32))
 
     jobs = orm.relationship("Job", cascade="all,delete", backref="analysis")
     delivery = orm.relationship("Delivery", uselist=False)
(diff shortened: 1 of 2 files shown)


trailblazer/store/models.py
```diff
@@ -92,6 +92,7 @@ class Analysis(Model):
     version = Column(types.String(32))
     workflow = Column(types.String(32))
     workflow_manager = Column(types.Enum(*WorkflowManager.list()), default=WorkflowManager.SLURM)
+    tower_workflow_id = Column(types.String(32))
```
Vince-janv, line 95: Should we default to `None` here, or do you plan to set it for the SLURM-analyses as well?

Vince-janv (approved):
I'm all for removing filesystem interactions 🙌

trailblazer/store/models.py
seallard: https://github.com/Clinical-Genomics/trailblazer/pull/483/commits/d482bc7a2a8de5030714d5a5ed71995d51a77e94

PR #488, opened 2024-08-29
Remove unused code

seallard:


islean (approved):
Looks good as far as I can tell. But take it with a grain of salt - I did not look at the initial PR too much.

PR #489, opened 2024-08-30
Remove tower workflow file interactions

seallard:
Blocked by https://github.com/Clinical-Genomics/IT-issues/issues/874 for production

I missed some instances where we still interact with the workflow file instead of using the workflow id stored in the database on the analysis.

Tested in stage, everything works.

### Fixed

- Swapped out old workflow file interactions when handling nf_tower analyses.

## Steps to test
1. Queue workflow run with `cg workflow rnafusion start cuddlyhen` in stage.
2. Go to cigrid and cancel analysis
3. Verify that the analysis is cancelled in tower as well https://cg-tower-stage.scilifelab.se

https://github.com/user-attachments/assets/67474cf7-66fd-496e-8001-8f41fb361cb5

### DIFF
--- a/trailblazer/services/tower/tower_api_service.py
+++ b/trailblazer/services/tower/tower_api_service.py
@@ -2,7 +2,7 @@ from trailblazer.clients.tower.models import TowerTasksResponse
 from trailblazer.clients.tower.tower_client import TowerAPIClient
 from trailblazer.constants import TOWER_WORKFLOW_STATUS, TrailblazerStatus
 from trailblazer.services.tower.error_handler import handle_errors
-from trailblazer.services.tower.utils import create_job_from_tower_task, get_tower_workflow_id
+from trailblazer.services.tower.utils import create_job_from_tower_task
 from trailblazer.store.models import Analysis, Job
 from trailblazer.store.store import Store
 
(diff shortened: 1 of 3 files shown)


trailblazer/services/tower/tower_api_service.py
```diff
@@ -17,8 +17,7 @@ def __init__(self, client: TowerAPIClient, store: Store) -> None:
     @handle_errors
     def update_jobs(self, analysis_id: int) -> list[Job]:
         analysis: Analysis = self.store.get_analysis_with_id(analysis_id)
-        workflow_id: str = get_tower_workflow_id(analysis)
-        response: TowerTasksResponse = self.client.get_tasks(workflow_id)
+        response: TowerTasksResponse = self.client.get_tasks(analysis.tower_workflow_id)
```
islean, line 20: Are the values in the column even populated yet? Seems like nothing has a tower_workflow_id yet?

seallard: They are populated, just not in stage :)

islean: I mean, I am checking in production

seallard: Weird! Looking into it. I'm guessing those analyses have not been run via the tower API which is why they have not been assigned a workflow id

seallard: Ah, I have forgotten to deploy the trailblazer web API which meant that the workflow id passed from cg was just dropped and never stored 😅

islean:
Tested again:

image

After cancelling:

image

Cancelled in tower:

image

trailblazer/services/tower/tower_api_service.py
islean: They do seem to be populating now.

islean (approved):
💯

islean:
Deployed to the VMs

PR #490, opened 2024-09-02
Add is_hidden parameter

islean:
## Description

Part of https://github.com/Clinical-Genomics/cigrid-ui/issues/493 and adds the option to hide a created analysis.

### Added

- Option to mark analysis as hidden upon creation.

### Changed

-

### Fixed

-

### How to prepare for test
- [ ] activate stage: `us`
- [ ] install on stage:
    ```Shell
    ```
- [ ] install on stage:

### How to test
- [ ] login to ...
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] tests executed by

### DIFF
--- a/trailblazer/store/crud/create.py
+++ b/trailblazer/store/crud/create.py
@@ -29,6 +29,7 @@ class CreateHandler(BaseHandler):
             type=analysis_data.type,
             workflow_manager=analysis_data.workflow_manager,
             tower_workflow_id=analysis_data.tower_workflow_id,
+            is_visible=not analysis_data.is_hidden,
         )
 
         if analysis_data.email:
--- a/tests/integration/services/analysis_service/test_analysis_service.py
+++ b/tests/integration/services/analysis_service/test_analysis_service.py
@@ -11,7 +11,6 @@ def test_updating_tower_analysis(
     tower_tasks_response: TowerTasksResponse,
     tower_workflow_response: TowerWorkflowResponse,
 ):
-
     # GIVEN an analysis started with tower without any job entries but running tasks
     analysis_service.job_service.tower_service.client.get_tasks.return_value = tower_tasks_response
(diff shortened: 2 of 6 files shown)


trailblazer/store/crud/create.py
```diff
@@ -29,6 +29,7 @@ def add_pending_analysis(self, analysis_data: CreateAnalysisRequest) -> Analysis
             type=analysis_data.type,
             workflow_manager=analysis_data.workflow_manager,
             tower_workflow_id=analysis_data.tower_workflow_id,
+            is_visible=not analysis_data.is_hidden,
```
diitaz93, line 32: why not using variables meaning the same thing both sites? It would be clearer if you named this parameter `is_hidden` instead of `is_visible`, as the one in the `CreateAnalysisRequest` is called `is_hidden`.

diitaz93 (approved):
👍 Just a small comment about the name, feel free to disregard

trailblazer/store/crud/create.py
islean: In Trailblazer the column is called `is_visible` so the request's parameter should be called `is_visible`. But the only time we actually want to specify this parameter is when we want the analysis hidden. Maybe it would be clearer to name it `is_visible` but it would also complicate the default behaviour a bit, as `is_visible == True` should be handled the same way as `is_visible == None`.

islean: I mean, the database model has the columned named `is_visible` so the only thing that could be changed is the `CreateAnalysisRequest` object. But the value will pretty much only be passed when you want to hide analyses. But maybe it would be clearer if it was named `is_visible` there as well?

diitaz93: ok it seems to extensive and out of scope, resolving

islean:
Deployed to the VMs

